Issue #22, opened 2019-04-30
Run tests against PyPy and Python 3 on Travis


Zac-HD:
After #20 is merged, it will be possible to run `futurize --stage2` in CI, then run the test suite against Python 3.  

(In our quick experiments on Monday, stage 2 changed the behaviour of `exec` and broke the tests on Python 2.  We'll want to fix that before the tutorial in June, but we can start testing against Python 3 even before we adopt those changes on master.)

Once we're running multiple Python versions in CI, it might also be useful to test against PyPy as well as Python 3 - even without the customized Bits support it should work as that's "just" a performance optimization.  Perhaps we should even test on PyPy3!

Zac-HD:
Closed by #37.

PR #37, opened 2019-05-27
Run tests on multiple Python versions

Zac-HD:
A quick approach to #22:  we can run the tests under Python 3 on Travis by inserting a transformer step, without (yet) requiring the code to be Python-3 compatible.

I've also added a `pypy` job, to catch any regressions on that side.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,4 +1,5 @@
 [settings]
+known_third_party = greenlet
 add_imports = from __future__ import absolute_import, from __future__ import division, from __future__ import print_function
 multi_line_output=3
 include_trailing_comma=True
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,6 @@ python:
  - "3.7"
  - "pypy3.5"
 
-matrix:
-  fast_finish: true
-  allow_failures:
-    - python: "3.6"
-    - python: "3.7"
-    - python: "pypy3.5"
-
 env:
  - CONFIG=""
  #- CONFIG="--test-verilog"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,21 @@
 # language and build matrix
 #------------------------------------------------------------------------
 
+dist: xenial
 language: python
 python:
  - "2.7"
+ - "pypy"
+ - "3.6"
+ - "3.7"
+ - "pypy3.5"
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - python: "3.6"
+    - python: "3.7"
+    - python: "pypy3.5"
 
 env:
  - CONFIG=""
@@ -43,7 +55,7 @@ install:
 #------------------------------------------------------------------------
 
 script:
- - futurize --write --stage1 .
+ - futurize --write --stage1 . &> /dev/null
  # - futurize --write --stage2 .
  - autoflake --in-place --remove-duplicate-keys $(find . -name '*.py')
  - pyupgrade --keep-percent-format $(find . -name '*.py')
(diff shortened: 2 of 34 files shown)


ptpan:
I checked the request history of Travis but unfortunately only the build requests from the last 24 hours are there. Several possibilities exist but I'm not sure which one leads to this issue... Maybe you can try to commit again and I'll check the build request?

> If a build hasn’t been triggered for your commit, these are the possible build request messages:
“Could not authorize build request”, usually means that the account’s subscription expired or that it ran out of trial builds.
“Build skipped via commit message”, this commit contains the skip command.
“GitHub payload is missing a merge commit”, please confirm your pull request is open and mergeable.
“Branch excluded per configuration” or “Branch not included per configuration”, please make sure your branch is not explicitly excluded or not included in your .travis.yml file.
Build type disabled via repository settings, please make sure your Push and Pull Request builds are still active.

Zac-HD:
OK! It turns out that certain syntax errors (I *think* omitting the quotes in python versions?) makes Travis silently ignore the whole branch.  Anyway, it's running now.

----------------

The good news is that [only four tests fail under PyPy](https://travis-ci.com/cornell-brg/pymtl3/jobs/203884512).

The less-good news is that Python 3 support won't be completely effortless.  I don't think it's too bad though; the ~100 failing tests remaining look mostly like I need to sprinkle some `__hash__` methods on the various classes and then look at how the `ast` module has changed.

So this remains a work in progress, but it *is* in progress!

jsn1993:
> OK! It turns out that certain syntax errors (I _think_ omitting the quotes in python versions?) makes Travis silently ignore the whole branch. Anyway, it's running now.
> 
> The good news is that [only four tests fail under PyPy](https://travis-ci.com/cornell-brg/pymtl3/jobs/203884512).
> 
> The less-good news is that Python 3 support won't be completely effortless. I don't think it's too bad though; the ~100 failing tests remaining look mostly like I need to sprinkle some `__hash__` methods on the various classes and then look at how the `ast` module has changed.
> 
> So this remains a work in progress, but it _is_ in progress!

Hey Zac if you merge in the latest commit from master, all tests should pass under pypy

jsn1993:
Also @Zac-HD I think the reason why there are a lot of hashable errors is that we kind of throw everything into a set/dict. For example I sometimes collect a bunch of functions into a set() and use them as keys in dictionaries. It worked fine in python2 but I'm not sure what happened in python3.

Zac-HD:
Nothing wrong with that pattern! It's just that custom objects aren't implicitly hashable in python 3, so we need to add the hash method.

jsn1993:
> Nothing wrong with that pattern! It's just that custom objects aren't implicitly hashable in python 3, so we need to add the hash method.

Hmmm really? I just tried to hash some Connectable and function in python3.6 it worked fine. What exactly are the new rules on custom objects?

Zac-HD:
From memory, instances of custom types are hashable by default, but *not* if you define an `__eq__` method (because "[hashable objects which compare equal must have the same hash value](https://docs.python.org/3/glossary.html)", and so the default `id()`-based implementation is not valid).

----------------------

Anyway!  I think this branch has grown enough, and is ready to be merged.

- Travis must pass on both CPython 2.7 and PyPy
- Travis *runs* against CPython 3.6, 3.7, and PyPy 3.5; but allows them to fail for now
- Improved, though unfinished, Python 3 support (37 tests failed, 461 passed)

So this is a clear improvement on the status quo.  

@cbatten - after merging, I'd be happy to keep working on Py3 support but I don't think I have a strong advantage for the remaining issues.  Let me know if you'd like me to keep going, or leave it to someone else.

Zac-HD:
Now down to just seven tests that fail on Python 3!  I've hit diminishing returns, but this is very close to working :smile:

Zac-HD:
Ready to merge!

PR #39, opened 2019-05-28
set_param

yo96:
More sophisticated `set_param` implementation.


pymtl3/dsl/NamedObject.py
```diff
@@ -55,13 +56,15 @@ def _construct( s ):
...
-      if "elaborate" in s._dsl.param_dict:
-        more_args = s._dsl.param_dict[ "elaborate" ].iteritems()
+      if "construct" in s._dsl.param_dict:
+        more_args = s._dsl.param_dict[ "construct" ].iteritems()
         kwargs.update( { x: y for x, y in more_args if x } )
 
       s.construct( *s._dsl.args, **kwargs )
 
       s._dsl.constructed = True
+      print( "="*30, "constrcut", "="*30 )
```
jsn1993: clean this up

pymtl3/passes/test/LineTraceParamPass.py
```diff
@@ -0,0 +1,48 @@
+"""
+#=========================================================================
+# MultiLevelLineTracePass_test.py
```
jsn1993: rename

pymtl3/passes/LineTraceParamPass.py
```diff
@@ -0,0 +1,39 @@
...
+
+from .BasePass import BasePass, PassMetadata
+
+
+class LineTraceParamPass( BasePass ):
+
+  def __call__( self, top ):
+
+    def wrap_line_trace( obj ):
+      if not hasattr( obj, '_ml_trace' ):
```
jsn1993: You might want to assert if you haven't applied this pass already.

pymtl3/dsl/NamedObject.py
```diff
@@ -72,25 +75,41 @@ def __setattr_for_elaborate__( s, name, obj ):
...
+            #     s._dsl.param_dict[ None ].iteritems():
+            #   if compiled_re.match( u_name ):
+            #     for x, y in subdict.iteritems(): # to merge two None subdicts
+            #       if x is None:
+            #         u._dsl.param_dict[ None ].update( subdict )
+            #       else:
+            #         u._dsl.param_dict[ x ] = y
+            
+            # Iterate through the param_dict and update u's dict
+            for key, subdict in s._dsl.param_dict.iteritems():
```
jsn1993: Are you going to change this to the new algorithm we discussed yesterday?

yo96: Both this and the new algorithm we discussed yesterday still won't work for some corner cases... I just re-implemented the whole thing again.

pymtl3/dsl/test/set_param_test.py
```diff
@@ -0,0 +1,118 @@
...
+  Z = Zoo()
+  Z.set_param( "top.construct", AnimalHouseTypes=[ PandaHouse, TigerTerrace, AnimalHouse, Aquarium ] )
+  Z.set_param( "top.houses[0].construct", AnimalTypes=[ Panda, Panda, Panda   ] )
+  Z.set_param( "top.houses[1].construct", AnimalTypes=[ Tiger, BabyTiger      ] )
+  Z.set_param( "top.houses[2].construct", AnimalTypes=[ Dromaius, HoneyBadger ] )
+  Z.set_param( "top.houses[1].animals[0].construct", MeatTypes=[ Dromaius, HoneyBadger ] )
+  Z.set_param( "top.houses[0].animals*.construct", lunch="bamboo", dinner="bamboo" )
+  Z.set_param( "top.houses[1].animals[0].meats*.construct", lunch="bamboo" )
+  Z.set_param( "top.houses[1].animals[0].meats[1].construct", lunch="grass" )
+  # The following set_param test tricky overwrite behavior.
```
yo96: This breaks the algorithm we discussed yesterday @jsn1993

yo96:
Cheng and I will add more tests and hopefully some microbenchmarking to evaluate how much the parametrization system slows down elaboration.

pymtl3/dsl/test/set_param_test.py
yo96: line 75~77

jsn1993: hmm why? the reverse orderof ordededdict still works for each parameter

jsn1993: hmmm let me think about it. When there are multiple parameters it's not only the name but also the argument name that makes a difference.

jsn1993:
> Cheng and I will add more tests and hopefully some microbenchmarking to evaluate how much the parametrization system slows down elaboration.

sounds good

pymtl3/dsl/test/set_param_test.py
yo96: yes exactly... for line 76, if I delete animals[0] and re-append the new dict, the 'dinner' would be lost
If i pop the node, merge it with the new dict, then re-append it to the end, that still won't work because it screws the order, since that means line 74 may overwrite line 75.

yo96: In my current algorithm, when inserting a node with '*', it will updates all previous nodes who matches the regex.
It's definitely slower though. I'm trying to do some microbenchmarking to see how much it may slow down the elaboration process.

pymtl3/passes/rtlir/rtype/RTLIRType.py
```diff
@@ -257,10 +257,12 @@ def _gen_parameters( s, obj ):
     arg_names = s.argspec.args[1:]
     assert s.argspec.varargs is None, "varargs are not allowed for construct!"
     assert s.argspec.keywords is None, "keyword args are not allowed for construct!"
-    kwargs = ()
+    # Yanghui : I commented out param_dict related stuff, since we are not
+    # adding extra key word arguments with set_param.
+    # kwargs = ()
```
yo96: @ptpan I don't think this part is necessary since `set_param` can only set the value for existing keyword arguments. It won't add extra constructor arguments. Can you take a look at this?

pymtl3/passes/rtlir/rtype/test/RTLIRDataType_test.py
```diff
@@ -31,6 +31,7 @@ def test_py_list():
   with expected_failure( RTLIRConversionError, 'should be a field of some struct' ):
     rdt.get_rtlir_dtype( [ 1, 2, 3 ] )
 
+# Yanghui : this test no longer passes.
 def test_py_struct_arg_no_default_value():
   class B( object ):
     def __init__( s, foo ):
```
yo96: @ptpan

pymtl3/passes/rtlir/rtype/RTLIRType.py
ptpan: Right these are not needed since we can rely on `_dsl.kwargs` to get all keyword argument values.

pymtl3/passes/rtlir/rtype/test/RTLIRDataType_test.py
ptpan: This branch improves the instantiation of a type and DSL no longer throws the exception this test originally expected. I've updated the test to match the new error message thrown from the RTLIR generation method.

yo96:
@jsn1993 I have finished everything I wanted to do with this PR. Please take a final look.

pymtl3/dsl/NamedObject.py
```diff
@@ -15,19 +15,136 @@
...
 
 class DSLMetadata(object):
   pass
 
+# NOTE: We found that the built-in OrderedDict slows down the elaboration
+# time because much time was spent calling OrderedDict.__init__. 
+# The time for instantiating an OrderedDict is quite long compared
+# to other primitive data structures such as list or dict. We have to
+# implement our own ordered dictionary to mitigate this overhead.
+# TODO: When we move to python3, we won't need this any more.
```
jsn1993: great

pymtl3/dsl/NamedObject.py
```diff
@@ -197,16 +307,16 @@ def set_parameter( s, string, value ):
   def _elaborate_construct( s ):
 
     if s._dsl.constructed:
-      print("Don't elaborate the same model twice. \
-             Use APIs to mutate the model.")
+      warnings.warn( "Don't elaborate the same model twice. "
```
jsn1993: Interesting. Peitian already mentioned this warning to me. We can discuss whether we want to have warnings throughout this project later. It's fine for now,

pymtl3/dsl/test/set_param_test.py
```diff
@@ -0,0 +1,178 @@
...
+  A.in_ = 1
+  A.tick()
+  print( "A.out==", A.out )
+  assert A.out == 3
+
+  B = IncrArray()
+  B.set_param( "top.construct", DataType=Bits8, num_incrs=3 )
+  B.set_param( "top.incrs*.construct", incr_value=3 )
+  B.set_param( "top.incrs[2].construct", incr_value=5 )
+  B.apply( SimpleSim )
```
jsn1993: Can you use simple_sim_pass here? We don't want to import stuff from passes/ in dsl.

PR #41, opened 2019-05-30
SystemVerilog translators

ptpan:
Implement SystemVerilog translators. The following constructs are name-mangled:
   - an array of components
   - an array of interfaces
   - all signals inside an interface

The implementation in this pull request uses the following SystemVerilog features:
   - struct to implement user defined structs
   - typedef to match the python struct name with the SystemVerilog struct name
   - always_comb and always_ff
   - logic

TODO:
   - Support nested interfaces
   - Add PyMTL upblk source code to translated .sv


jsn1993:
Are we going to support an array of components in a future PR or it's not in the plan?

ptpan:
> Are we going to support an array of components in a future PR or it's not in the plan?

An array of components is supported -- just replicate all components

jsn1993:
> > Are we going to support an array of components in a future PR or it's not in the plan?
> 
> An array of components is supported -- just replicate all components

Ok. I think I wasn't clear in the original question. Do you plan to generate a list of components in SV instead of name mangling them?

pymtl3/passes/sverilog/translation/TranslationPass.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import absolute_import, division, print_function
+
+import os
+
+from pymtl3.passes.BasePass import BasePass, PassMetadata
+
+from .SVTranslator import SVTranslator
+
+
+def mk_TranslationPass( _SVTranslator ):
```
jsn1993: Instead of directly using SVTranslator in the pass, why do you need this function to make a pass?

ptpan:
> > > Are we going to support an array of components in a future PR or it's not in the plan?
> > 
> > 
> > An array of components is supported -- just replicate all components
> 
> Ok. I think I wasn't clear in the original question. For a list of pymtl3 components, do you plan to generate an array of components in SV instead of name mangling the pymtl3 components into a bunch of SV XXX_0 XXX_1 modules?

The current plan is to name mangle that python list of components.

pymtl3/passes/sverilog/translation/TranslationPass.py
ptpan: For testing purposes just like what we did for translators (I'll add the test cases after the import pass PR). I should have added more comments here...

jsn1993:
> > > > Are we going to support an array of components in a future PR or it's not in the plan?
> > > 
> > > 
> > > An array of components is supported -- just replicate all components
> > 
> > 
> > Ok. I think I wasn't clear in the original question. For a list of pymtl3 components, do you plan to generate an array of components in SV instead of name mangling the pymtl3 components into a bunch of SV XXX_0 XXX_1 modules?
> 
> The current plan is to name mangle that python list of components.

But eventually?

ptpan:
> > > > > Are we going to support an array of components in a future PR or it's not in the plan?
> > > > 
> > > > 
> > > > An array of components is supported -- just replicate all components
> > > 
> > > 
> > > Ok. I think I wasn't clear in the original question. For a list of pymtl3 components, do you plan to generate an array of components in SV instead of name mangling the pymtl3 components into a bunch of SV XXX_0 XXX_1 modules?
> > 
> > 
> > The current plan is to name mangle that python list of components.
> 
> But eventually?

Eventually we want to use `generate` statements. The tricky part of that is all index operations on an array of components should be transformed into index operations on an array of wires, which is the reason why we should implement that later...

jsn1993:
@ptpan hmmm do you think every backend should be implemented in the same way as your SV backend which contains different levels corresponding to RTLIR level?

ptpan:
> @ptpan hmmm do you think every backend should be implemented in the same way as your SV backend which contains different levels corresponding to RTLIR level?

No that is not necessary -- we did this because it helps testing and matches our development narratives. 

If the backend is relatively simple it is fine to just have all callback methods in a single file. On the other hand if someone is trying to implement a complicated backend and he/she finds our development narrative helpful, he/she might also adopt our approach.

jsn1993:
So maybe we should develop the yosys backend in that way?

ptpan:
> So maybe we should develop the yosys backend in that way?

I'm not 100% sure about that yet... We are already name-mangling interfaces and components in SystemVerilog backend which means for yosys we only need to add array and struct name mangle support. If the new part is not too large we might as well overwrite a few methods in a single class.

PR #44, opened 2019-06-04
Placeholder and replace_component

jsn1993:
Stuff in this PR:
1. A major refactoring over the existing dsl code base to centralize all API implementations in Component. 
2. Implement Placeholder in dsl and add support in elaboration to recognize Placeholders and throw error in passes if there are leftover placeholders.
3. I started off implementing the delete_placeholder API but I realize that the delete_component is not that different lol. At the end I refactored add and delete into private functions and just expose replace_component to outside world.

### DIFF
--- a/pymtl3/dsl/Component.py
+++ b/pymtl3/dsl/Component.py
@@ -91,6 +91,112 @@ class Component( ComponentLevel7 ):
       s._dsl.all_method_nets = s._resolve_method_connections()
       s._dsl._has_pending_connections = False
 
+  # Override
+  def _add_component( s, parent, name, indices, obj, provided_connections=[] ):
+    try:
+      top = s._dsl.elaborate_top
+    except AttributeError:
+      raise NotElaboratedError()
+
+    # Check if we are adding obj to a list of to a component
+    if not indices:
+      # If we are adding field s.x, we simply reuse the setattr hook
+      assert not hasattr( s, name )
+      NamedObject.__setattr__ = NamedObject.__setattr_for_elaborate__
+      setattr( parent, name, obj )
+      del NamedObject.__setattr__
+
+    else:
+      assert hasattr( s, name )
+
+      # We use the indices to get the list parent and set the element
+
+      list_parent = getattr( parent, name )
+      i = 0
+      while i < len(indices) - 1:
+        list_parent = list_parent[ indices[i] ]
+        i += 1
+      list_parent[ indices[i] ] = obj
+
+      # NOTE THAT we cannot use setattr hook for the top level metadata
+      # because setting an element in the list doesn't call setattr.
+      # As a result we have to copy some code from setattr hook and do it
+      # manually here for the top level
+
+      obj._dsl.parent_obj = parent
+      obj._dsl.level      = parent._dsl.level + 1
+      obj._dsl.my_name    = u_name = name + "".join( [ "[{}]".format(x)
+                                                     for x in indices ] )
+      # FIXME properly implement param_tree
+      # Iterate through the param_tree and update u
+      if parent._dsl.param_tree is not None:
+        if parent._dsl.param_tree.children is not None:
+          for comp_name, node in parent._dsl.param_tree.children.iteritems():
+            if comp_name == u_name:
+              # Lazily create the param tree
+              if obj._dsl.param_tree is None:
+                obj._dsl.param_tree = ParamTreeNode()
+              obj._dsl.param_tree.merge( node )
+
+            elif node.compiled_re is not None:
+              if node.compiled_re.match( u_name ):
+                # Lazily create the param tree
+                if obj._dsl.param_tree is None:
+                  obj._dsl.param_tree = ParamTreeNode()
+                obj._dsl.param_tree.merge( node )
+
+      obj._dsl.full_name = ( parent._dsl.full_name + "." + u_name )
+
+      # store the name/indices
+      obj._dsl._my_name     = name
+      obj._dsl._my_indices  = indices
+
+      NamedObject.__setattr__ = NamedObject.__setattr_for_elaborate__
+      obj._construct()
+      del NamedObject.__setattr__
+
+    top = s._dsl.elaborate_top
+
+    added_components = obj._collect_all( [ lambda x: isinstance( x, Component ) ] )[0]
+
+    # First elaborate all functions to spawn more named objects
+    for c in added_components:
+      c._dsl.elaborate_top = top
+      c._elaborate_read_write_func()
+
+    added_signals, added_method_ports = \
+      obj._collect_all( [ lambda x: isinstance( x, Signal ), \
+                          lambda x: isinstance( x, MethodPort ) ] )
+
+    top._dsl.all_components    |= added_components
+    top._dsl.all_signals       |= added_signals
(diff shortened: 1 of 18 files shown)


jsn1993:
I finished handling method ports and interfaces. Replacing a placeholder with a queue and let the queue take a parameter to configure himself to bypass/pipe looks very neat.

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,542 @@ def _construct( s ):
...
+
+    for c in added_components:
+      top._collect_vars( c )
+
+    for c in added_signals | added_method_ports:
+      c._dsl.elaborate_top = top
+
+    # Lazy -- to avoid resolve_connection call which takes non-trivial
+    # time upon adding any connect, I just mark it here. Please make sure
+    # to call s.get_all_value_nets() to flush all pending connections
```
ptpan: I think the pending connections will be resolved once you need it right? The comment here sounds like your have to manually call some API to resolve connections before using them...

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,543 @@ def _construct( s ):
...
+
+    return _delete_component_internal( top, obj )
+    # import gc
+    # gc.collect() # this takes 0.1 seconds
+
+  #-----------------------------------------------------------------------
+  # Post-elaborate public APIs (can only be called after elaboration)
+  #-----------------------------------------------------------------------
+
+  """ Convenience/utility APIs """
```
ptpan: I think docstring generally either appears right after a function/class definition or at the beginning of a module? Plus this doctoring will not be collected into `__doc__` of any methods...

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,543 @@ def _construct( s ):
...
+        if is_list:
+          s._dsl.swapped_values[ component ] = ( current_obj, i, current_obj[i], is_list )
+          current_obj[i] = obj
+        else:
+          s._dsl.swapped_values[ component ] = ( current_obj, i, getattr(current_obj, i), is_list )
+          setattr( current_obj, i, obj )
+
+    s._dsl.locked_simulation = False
+
+  """ APIs that provide local metadata of a component """
```
ptpan: Same here. I feel our `#-----` style separator works well in these kinds of situations. What do you think?

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,543 @@ def _construct( s ):
...
+        iterable = enumerate( current_obj )
+        is_list = True
+      elif isinstance( current_obj, NamedObject ):
+        iterable = current_obj.__dict__.iteritems()
+        is_list = False
+      else:
+        return
+
+      for i, obj in iterable:
+        if not is_list and i.startswith("_"): # impossible to have tuple
```
ptpan: `is_list` is constant within this for loop so you probably want to move the its check outside `for` to have better performance.

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,543 @@ def _construct( s ):
...
+      parent._dsl.connect_order = new_connect_order
+
+      return saved_connections
+
+    return _delete_component_internal( top, obj )
+    # import gc
+    # gc.collect() # this takes 0.1 seconds
+
+  #-----------------------------------------------------------------------
+  # Post-elaborate public APIs (can only be called after elaboration)
```
ptpan: I like the new API organization so much -- you don't have to dig into different levels to find out them!

pymtl3/dsl/Component.py
```diff
@@ -55,10 +61,543 @@ def _construct( s ):
...
+    # call garbage collector. If we can make sure it is collected
+    # automatically and fast enough, we might remove this extra layer
+    #
+    # EDIT: After experimented w/ and w/o gc.collect(), it seems like it
+    # is eventually collected, but sometimes the intermediate memory
+    # footprint might reach up to gigabytes, so let's keep the
+    # gc.collect() for now
+
+    def _delete_component_internal( top, foo ):
+      is_placeholder = isinstance( obj, Placeholder )
```
ptpan: Seems like you are not assuming `Placeholder` here.  What's this flag for?

pymtl3/dsl/ComponentLevel2.py
```diff
@@ -451,7 +464,7 @@ def add_constraints( s, *args ): # add RD-U/WR-U constraints
           s._dsl.WR_U_constraints[ x0.var ].add( (sign, x1.func) )
 
   #-----------------------------------------------------------------------
-  # elaborate
+  # elaborate-related public APIs
```
ptpan: Did you mean just `API` instead of `public API`?

pymtl3/dsl/Component.py
jsn1993: Good point. The comment is copied from the old implementation. What I actually mean is that whenever you call get_all_value_nets which is the API that gives you all the nets, the API itself will check if there are pending connections and flush them to return the latest nets.

pymtl3/dsl/Component.py
jsn1993: hmmm then what is a good seperator of blocks of APIs?

ptpan: What do you think of `#--` here?

pymtl3/dsl/Component.py
jsn1993: Hmmm good catch. There is one line below that checks isinstance(obj,Placeholder) again. The purpose was to check if it is placeholder. If it is, I don't need to call those uncollect thingy.

pymtl3/dsl/ComponentLevel2.py
jsn1993: I guess I copy-paste a lot and now I need to fix them ...

pymtl3/dsl/Component.py
jsn1993: I want to make those "subsections" of the "Post-elaborate public APIs" section. In line 321-323 I already use #--. Any idea on a different one?

ptpan: How about a single line `# Convenience/utility APIs` with a blank line before/after it? Or `## Convenience/utility APIs` like in markdown?

pymtl3/dsl/test/Placeholder_test.py
```diff
@@ -0,0 +1,299 @@
...
+    def construct( s ):
+      s.q = deque()
+      s.add_constraints( M(s.enq) < M(s.deq) )
+
+  a = FooCL_top()
+  a.elaborate()
+
+  a.replace_component( a.x.foo, RealQueue )
+
+  simple_sim_pass( a )
```
ptpan: Is it possible to have some preliminary assertions here? Just to make sure this works as intended.

jsn1993: hmm will do

pymtl3/dsl/Component.py
jsn1993: Well if it's single line, we have too many single line comments for each function in that part  ...

jsn1993: @ptpan I guess we should spend sometime later to clean up all comment formats in one shot ...

PR #46, opened 2019-06-07
Kernel bug fix: explicit reset signal, disconnecting consts, connecting Bits to signal

jsn1993:
This PR fixed three bugs:

1. We connect the parent clk/reset to implicit reset signal before we call construct on the component. However, when the user declares a reset/clk signal in construct block, it overrides the previously declared signals, and we will instead collect the new signal later, but the connection is with the old signal. To solve this we postpone connecting the clk/reset to parent after we call construct (note that we cannot move the declaration below construct because the user may use the reset signal inside the construct block).

2. Previously I didn't handle the case where a constant connection needs to be saved when you delete a component.

3.I implemented connecting a Bits/int object to a Bits/int signal and add tests to catch the failed cases. Currently we hard-code int/Bits -- we allow an int const to be connected to int/Bits signals, and a Bits const to be connected to Bits signals **with the same bitwidth**. We throw an error when we connect Bits/int to a struct signal. In the future, I would imagine we might be able to connect a Bits constant to a Bitstruct using the pack/unpack facility. We might also want to connect a struct const to a struct signal.

### DIFF
--- a/pymtl3/dsl/Component.py
+++ b/pymtl3/dsl/Component.py
@@ -51,12 +51,10 @@ class Component( ComponentLevel7 ):
       # We hook up the added clk and reset signals here. NOTE THAT if the
       # user overwrites clk/reset inside the component, we still get the
       # correct connection.
-      try:
-        parent = s.get_parent_object()
+      parent = s.get_parent_object()
+      if parent is not None:
         parent.connect( s.clk, parent.clk )
         parent.connect( s.reset, parent.reset )
-      except AttributeError:
-        pass
 
       if s._dsl.call_kwargs is not None: # s.a = A()( b = s.b )
         s._continue_call_connect()
--- a/pymtl3/dsl/Component.py
+++ b/pymtl3/dsl/Component.py
@@ -15,7 +15,7 @@ from pymtl3.datatypes import Bits1
 
 from .ComponentLevel1 import ComponentLevel1
 from .ComponentLevel7 import ComponentLevel7
-from .Connectable import InPort, Interface, MethodPort, OutPort, Signal, Wire
+from .Connectable import Const, InPort, Interface, MethodPort, OutPort, Signal, Wire
 from .errors import InvalidAPICallError, InvalidConnectionError, NotElaboratedError
 from .NamedObject import NamedObject
 from .Placeholder import Placeholder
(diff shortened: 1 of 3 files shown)


pymtl3/dsl/Component.py
```diff
@@ -45,17 +45,19 @@ def _construct( s ):
...
-      # We hook up the added clk and reset signals here.
+      # Same as parent class _construct
+      s.construct( *s._dsl.args, **kwargs )
+
+      # We hook up the added clk and reset signals here. NOTE THAT if the
+      # user overwrites clk/reset inside the component, we still get the
+      # correct connection.
       try:
         parent = s.get_parent_object()
         parent.connect( s.clk, parent.clk )
```
ptpan: is it possible that some statements inside `s.connect` throws an attribute error?

ptpan:
I remember after elaboration, `Const` objects will not have `_dsl.my_name` attribute. Is there any specific reason for this? Just curious...

pymtl3/dsl/Component.py
ptpan: I had some situations where I use try-except to catch some attribute error but unintentionally missed exceptions thrown from the function calls inside the block.

jsn1993: oh good catch! Do I actually need this try-catch block here? I think the reason I had this was when I instantiate a ComponentLevel6 as a child model inside ComponentLevel3 and ComponengLevel3 doesn't have reset/clk. Now since we are exposing only Component to the user, I guess we don't need this anymore.

jsn1993:
> I remember after elaboration, `Const` objects will not have `_dsl.my_name` attribute. Is there any specific reason for this? Just curious...

This is because it's a constant .. The only scenario you create a Const object is when you do ```s.connect( s.x, 1 )```. What kind of name were you thinking about?

ptpan:
> > I remember after elaboration, `Const` objects will not have `_dsl.my_name` attribute. Is there any specific reason for this? Just curious...
> 
> This is because it's a constant .. The only scenario you create a Const object is when you do `s.connect( s.x, 1 )`. What kind of name were you thinking about?

If we have `s.STATE_IDLE = 0` and do `s.connect(s.state.in_, s.STATE_IDLE)`, the python integer `0` is actually an attribute of the component. During translation it is turned into a localparam but is not used because the translation pass cannot know its name from `_dsl`.

jsn1993:
> > > I remember after elaboration, `Const` objects will not have `_dsl.my_name` attribute. Is there any specific reason for this? Just curious...
> > 
> > 
> > This is because it's a constant .. The only scenario you create a Const object is when you do `s.connect( s.x, 1 )`. What kind of name were you thinking about?
> 
> If we have `s.STATE_IDLE = 0` and do `s.connect(s.state.in_, s.STATE_IDLE)`, the python integer `0` is actually an attribute of the component. During translation it is turned into a localparam but is not used because the translation pass cannot know its name from `_dsl`.

Well, I have no way to know the name of the signal inside s.connect(). The only thing it sees is a "0".

ptpan:
> > > > I remember after elaboration, `Const` objects will not have `_dsl.my_name` attribute. Is there any specific reason for this? Just curious...
> > > 
> > > 
> > > This is because it's a constant .. The only scenario you create a Const object is when you do `s.connect( s.x, 1 )`. What kind of name were you thinking about?
> > 
> > 
> > If we have `s.STATE_IDLE = 0` and do `s.connect(s.state.in_, s.STATE_IDLE)`, the python integer `0` is actually an attribute of the component. During translation it is turned into a localparam but is not used because the translation pass cannot know its name from `_dsl`.
> 
> Well, I have no way to know the name of the signal inside s.connect(). The only thing it sees is a "0".

I see.

PR #47, opened 2019-06-07
stdlib cleanup

yo96:
In this PR I added Type as constructor for CL test src/sink in order to provide better line trace. I also use Type to infer the line trace width for send/recv interfaces.


pymtl3/stdlib/ifcs/SendRecvIfc.py
```diff
@@ -28,7 +28,13 @@ def construct( s, Type ):
     s.MsgType = Type
 
   def line_trace( s ):
-    return enrdy_to_str( s.msg, s.en, s.rdy )
+    try:
+      trace_len = s.trace_len
+    except AttributeError:
+      s.trace_len = len( "{}".format( s.MsgType() ) )
```
jsn1993: this is correct -- it will throw exception for the first time and then always get the s.trace_len

PR #52, opened 2019-06-09
Run flake8 in CI

Zac-HD:
`flake8` is great for finding shallow bugs!  In this config we select only "fatal" errors, then ignore several of them too - but the remaining issues are real, and it'll defend against a variety of possible future issues.

[Here's the list of `flake8` error codes](http://flake8.pycqa.org/en/latest/user/error-codes.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,12 +50,10 @@ install:
 
 script:
  - futurize --write --stage1 ./pymtl3 &> /dev/null
- # - futurize --write --stage2 .
- - autoflake --in-place --remove-duplicate-keys $(find ./pymtl3 -name '*.py')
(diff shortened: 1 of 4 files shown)


jsn1993:
So flake8 has some overlap with autoflake, but also has some other features that complement the tools we already have?

Zac-HD:
flake8 can't fix your code, but can output lots of warnings about potential problems like unused imports or variables, redefined functions, scope issues, etc. In this case I've configured it to ignore code formatting entirely, and also ignore some PyMTL idioms. Concretely, even with such a restricted config it noticed "raise NotImplemented", which should use NotImplementedError instead. 

autoflake actually edits your code to fix a subset of the warnings produced by flake8, including unused standard library imports.

PR #53, opened 2019-06-09
RTL queues

yo96:
In this PR I implemented RTL pipe queue, bypass queue, and single-entry normal queue, pipe queue and bypass queues. All of them have enq/deq interface.

### DIFF
--- a/pymtl3/stdlib/rtl/queues.py
+++ b/pymtl3/stdlib/rtl/queues.py
@@ -510,7 +510,7 @@ class PipeQueue1EntryRTL( Component ):
       if s.reset:
         s.full = b1(0)
       else:
-        s.full = ~s.deq.en & (s.enq.en | s.full) | s.enq.en & s.full
+        s.full = s.enq.en | s.full & ~s.deq.en
 
     @s.update_on_edge
     def up_entry():
--- a/pymtl3/stdlib/rtl/queues.py
+++ b/pymtl3/stdlib/rtl/queues.py
@@ -55,11 +55,11 @@ class NormalQueueCtrlRTL( Component ):
 
     # Interface
 
-    s.enq_en  = InPort ( Bits1   )
-    s.enq_rdy = OutPort( Bits1   )
-    s.deq_en  = InPort ( Bits1   )
-    s.deq_rdy = OutPort( Bits1   )
-    s.count   = OutPort( PtrType )
+    s.enq_en  = InPort ( Bits1     )
+    s.enq_rdy = OutPort( Bits1     )
+    s.deq_en  = InPort ( Bits1     )
+    s.deq_rdy = OutPort( Bits1     )
+    s.count   = OutPort( CountType )
 
     s.wen     = OutPort( Bits1   )
     s.waddr   = OutPort( PtrType )
@@ -129,7 +129,7 @@ class NormalQueueRTL( Component ):
 
     s.enq   = EnqIfcRTL( EntryType )
     s.deq   = DeqIfcRTL( EntryType )
-    s.count = OutPort( mk_bits( clog2( num_entries ) ) )
+    s.count = OutPort( mk_bits( clog2( num_entries+1 ) ) )
 
     # Components
 
@@ -176,11 +176,11 @@ class PipeQueueCtrlRTL( Component ):
 
     # Interface
 
-    s.enq_en  = InPort ( Bits1   )
-    s.enq_rdy = OutPort( Bits1   )
-    s.deq_en  = InPort ( Bits1   )
-    s.deq_rdy = OutPort( Bits1   )
-    s.count   = OutPort( PtrType )
+    s.enq_en  = InPort ( Bits1     )
+    s.enq_rdy = OutPort( Bits1     )
+    s.deq_en  = InPort ( Bits1     )
+    s.deq_rdy = OutPort( Bits1     )
+    s.count   = OutPort( CountType )
 
     s.wen     = OutPort( Bits1   )
     s.waddr   = OutPort( PtrType )
@@ -256,7 +256,7 @@ class PipeQueueRTL( Component ):
 
     s.enq   = EnqIfcRTL( EntryType )
     s.deq   = DeqIfcRTL( EntryType )
-    s.count = OutPort( mk_bits( clog2( num_entries ) ) )
+    s.count = OutPort( mk_bits( clog2( num_entries+1 ) ) )
 
     # Components
 
@@ -332,11 +332,11 @@ class BypassQueueCtrlRTL( Component ):
 
     # Interface
 
-    s.enq_en  = InPort ( Bits1   )
-    s.enq_rdy = OutPort( Bits1   )
-    s.deq_en  = InPort ( Bits1   )
-    s.deq_rdy = OutPort( Bits1   )
-    s.count   = OutPort( PtrType )
+    s.enq_en  = InPort ( Bits1     )
+    s.enq_rdy = OutPort( Bits1     )
+    s.deq_en  = InPort ( Bits1     )
+    s.deq_rdy = OutPort( Bits1     )
+    s.count   = OutPort( CountType )
 
     s.wen     = OutPort( Bits1   )
     s.waddr   = OutPort( PtrType )
@@ -416,7 +416,7 @@ class BypassQueueRTL( Component ):
 
     s.enq   = EnqIfcRTL( EntryType )
     s.deq   = DeqIfcRTL( EntryType )
-    s.count = OutPort( mk_bits( clog2( num_entries ) ) )
+    s.count = OutPort( mk_bits( clog2( num_entries+1 ) ) )
 
     # Components
 
--- a/pymtl3/stdlib/rtl/queues.py
+++ b/pymtl3/stdlib/rtl/queues.py
@@ -216,10 +216,10 @@ class PipeQueueCtrlRTL( Component ):
 
     @s.update
     def up_enq_rdy():
-      if ~s.reset:
-        s.enq_rdy = ( s.count < s.num_entries ) | s.deq_en
-      else:
+      if s.reset:
         s.enq_rdy = b1(0)
+      else:
+        s.enq_rdy = ( s.count < s.num_entries ) | s.deq_en
 
 
     @s.update
@@ -378,10 +378,10 @@ class BypassQueueCtrlRTL( Component ):
 
     @s.update
     def up_deq_rdy():
-      if ~s.reset:
-        s.deq_rdy = ( s.count > CountType(0) ) | s.enq_en
-      else:
+      if s.reset:
         s.deq_rdy = b1(0)
+      else:
+        s.deq_rdy = ( s.count > CountType(0) ) | s.enq_en
     
     @s.update
     def up_mux_sel():
@@ -498,10 +498,10 @@ class NormalQueue1EntryRTL( Component ):
 
     @s.update
     def up_enq_rdy():
-      if ~s.reset:
-        s.enq.rdy = ~s.full
-      else:
+      if s.reset:
         s.enq.rdy = b1(0)
+      else:
+        s.enq.rdy = ~s.full
 
     @s.update
     def up_deq_rdy():
--- a/pymtl3/stdlib/rtl/queues.py
+++ b/pymtl3/stdlib/rtl/queues.py
@@ -89,7 +89,7 @@ class NormalQueueCtrlRTL( Component ):
     def up_rdy_signals():
       if ~s.reset:
         s.enq_rdy = s.count < s.num_entries
-        s.deq_rdy = s.count > 0
+        s.deq_rdy = s.count > CountType(0)
       else:
         s.enq_rdy = b1(0)
         s.deq_rdy = b1(0)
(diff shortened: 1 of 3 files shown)


pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+
+
+    @s.update
+    def up_xfer_signals():
+      s.enq_xfer  = s.enq_en and s.enq_rdy
+      s.deq_xfer  = s.deq_en and s.deq_rdy
+
+    @s.update
+    def up_next():
+      s.head_next = s.head - b1(1) if s.head > 0 else PtrType( s.last_idx )
```
jsn1993: do we need multi-entry pipe and bypass queues? let me bring this up on slack.

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +444,170 @@ def construct( s, MsgType, num_entries=2 ):
...
+        s.full = b1(1)
+      elif s.full and ~s.enq.en and s.deq.en:
+        s.full = b1(0)
+      else:
+        s.full = s.full
+    
+    # TODO: figure out whether to use deepcopy here.
+    @s.update_on_edge
+    def up_entry():
+      s.entry = s.enq.msg if s.enq.en and ~s.deq.en else s.entry
```
jsn1993: if we use deepcopy it's going to be ```if s.enq.en and ~s.deq.en: s.entry = deepcopy(s.enq.msg)```

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +444,170 @@ def construct( s, MsgType, num_entries=2 ):
...
+
+    @s.update
+    def up_enq_rdy():
+      if ~s.reset:
+        s.enq.rdy = b1(1) if ~s.full else b1(0)
```
jsn1993: ```s.enq.rdy = ~s.full``` ...

jsn1993: actually ... should we combinationally depend on reset signal? @cbatten

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +444,170 @@ def construct( s, MsgType, num_entries=2 ):
...
+      else:
+        s.enq.rdy = b1(0)
+
+    @s.update
+    def up_deq_rdy():
+      if ~s.reset:
```
jsn1993: same here

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +444,170 @@ def construct( s, MsgType, num_entries=2 ):
...
+    s.entry = Wire( EntryType )
+    s.full  = Wire( Bits1 )
+
+    # Logic
+
+    @s.update_on_edge
+    def up_full():
+      if s.reset:
+        s.full = b1(0)
+      elif ~s.full and s.enq.en and ~s.deq.en:
```
jsn1993: the simplified expression is  ```s.full  = ~s.deq.en & (s.enq.en | s.full) ```

jsn1993: well something like this is also fine .. I guess this will create more bridges for simulation though. Might as well just have simplified form in stdlib

pymtl3/stdlib/rtl/queues.py
yo96: we have these in cl, so I think we should have their RTL equivalence in our stdlib as well.

pymtl3/stdlib/rtl/queues.py
yo96: should we use deep copy here though? not sure

pymtl3/stdlib/rtl/queues.py
jsn1993: I was just thinking about the discussion between @cbatten and I on whether a pipe queue with >1 entry is useful. The conclusion was that >=2 entry normal queue is good enough for most cases to buffer the message without elongating the ready path, and we would usually just go for a big normal queue to cut both ready and valid. >=2 entry bypass queue's behavior is a little bit weird -- it can only bypass when the queue is empty. Otherwise the queue acts like a normal queue. We do use a 2-entry bypass queue in ece4750 processor to sustain 1-fetch per cycle during squashing.

pymtl3/stdlib/rtl/queues.py
jsn1993: I mean the general question you want to understand is, when do you need deepcopy? In this case, if s.enq.msg is a struct, when you assignit to s.entry without deepcopy, s.entry stores a pointer that points to the struct object. Let's say and this cycle your queue is full, so s.entry the struct object remain unchanged, and somehow some other update block in the producer unit change some field of this object (like s.imem.req.addr = pc+4). It will change the field of the object and unfortunately s.entry points to the same object if we don't do deepcopy. As a result s.entry is magically changed which is not what you want. This is one scenario. The other scenario is that s.entry points to a Bits object, and someone else do s.enq.msg[1:3] = 2. s.entry points to the same object without deepcopy.

pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+    # Connections
+
+    s.connect( s.wen,   s.enq_xfer )
+    s.connect( s.waddr, s.tail     )
+    s.connect( s.raddr, s.head     )
+
+    @s.update
+    def up_enq_rdy():
+      if ~s.reset:
+        s.enq_rdy = b1(1) if s.count < s.num_entries else b1(0)
```
jsn1993: how about ```s.enq_rdy = s.count < s.num_entries```?

pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+      else:
+        s.enq_rdy = b1(0)
+
+    @s.update
+    def up_deq_rdy():
+      if ~s.reset:
+        s.deq_rdy = b1(1) if s.count > 0 or s.enq_en else b1(0)
```
jsn1993: same here

pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+      else:
+        s.deq_rdy = b1(0)
+    
+    @s.update
+    def up_mux_sel():
+      s.mux_sel = b1(0) if s.count > 0 else b1(1)
```
jsn1993: same here

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +444,170 @@ def construct( s, MsgType, num_entries=2 ):
...
+        s.full = b1(1)
+      elif s.full and s.deq.en:
+        s.full = b1(0)
+      else:
+        s.full = s.full
+    
+    # TODO: figure out whether to use deepcopy here.
+    @s.update_on_edge
+    def up_entry():
+      s.entry = s.enq.msg if s.enq.en else s.entry
```
jsn1993: I don't like s.entry = x if y else s.entry ... It means you assign yourself to yourself

yo96: it's sequential logic, in verilog I would do s.entry <= y ? x : s.entry. Of course you should now do this in combinational logic.

pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+    s.connect( s.ctrl.waddr,   s.dpath.waddr   )
+    s.connect( s.ctrl.raddr,   s.dpath.raddr   )
+
+    # Connect to interface
+
+    s.connect( s.enq.en,  s.ctrl.enq_en   )
+    s.connect( s.enq.rdy, s.ctrl.enq_rdy  )
+    s.connect( s.deq.en,  s.ctrl.deq_en   )
+    s.connect( s.deq.rdy, s.ctrl.deq_rdy  )
+    s.connect( s.count,   s.ctrl.count    )
```
yo96: Just found a bug here. Bit width of `s.count` and `s.ctrl.count` does not match. Why isn't `s.connect` complaining? @jsn1993

yo96: Oh I see, sometimes their bit width can be the same. Our tests should test more configurations.

pymtl3/stdlib/rtl/queues.py
yo96: does boolean expression returns Bits1?

pymtl3/stdlib/rtl/queues.py
```diff
@@ -86,11 +86,298 @@ def construct( s, num_entries=2 ):
...
+    # Connections
+
+    s.connect( s.wen,   s.enq_xfer )
+    s.connect( s.waddr, s.tail     )
+    s.connect( s.raddr, s.head     )
+
+    @s.update
+    def up_rdy_signals():
+      if ~s.reset:
+        s.deq_rdy = b1(1) if s.count > 0 else b1(0)
```
yo96: Is it OK to use integer 0 here? @jsn1993

jsn1993: I think it's ok for simulation but not sure if the translation pass enforce it ...

pymtl3/stdlib/rtl/queues.py
jsn1993: I think yes according to verilog conventions and I implemented Bits in that way https://github.com/cornell-brg/pymtl3/blob/master/pymtl3/datatypes/PythonBits.py#L125

pymtl3/stdlib/rtl/queues.py
jsn1993: oh this is update on edge .. By the way you might want to use deepcopy here to deepcopy(s.enq.msg)

pymtl3/stdlib/rtl/queues.py
yo96: uhhh it's kind of inconvenient to worry about types even for 0 and 1...

pymtl3/stdlib/rtl/queues.py
yo96: Already did!

pymtl3/stdlib/rtl/queues.py
tancheng: I tried... need explicitly define the bitwidth for 0 here for translation (should be the same type as s.count)...

jsn1993: Hmmm @ptpan what do you think?

pymtl3/stdlib/rtl/queues.py
yo96: if you do k-map it's actually `s.full = ~s.deq.en & (s.enq.en | s.full) | s.enq.en & s.full` (or `s.full = s.enq.en & ~s.deq.en | s.enq.en & s.full | ~s.deq.en & s.full`).

yo96: for single entry bypass queue your expression is correct because enq.en and deq.en when s.full is a don't care.

pymtl3/stdlib/rtl/queues.py
```diff
@@ -89,17 +91,17 @@ def up_rdy_signals():
         s.enq_rdy = s.count < s.num_entries
```
jsn1993: how about ```s.enq_rdy = (s.count < s.num_entries) & ~s.reset```

yo96: alright

pymtl3/stdlib/rtl/queues.py
yo96: am I allowed to add bits with different width? like `s.head = s.head + b1(1)`?

pymtl3/stdlib/rtl/queues.py
yo96: should we prioritize performance or readability in our stdlib?

pymtl3/stdlib/rtl/queues.py
tancheng: current translation requires exactly the same bitwidth on each side.

cbatten:
@yo96 did you add a test that would have failed because of the bitwidth mismatch? remember if we find a bug you should always add a directed test that fails, then fix the bug

pymtl3/stdlib/rtl/queues.py
```diff
@@ -156,3 +436,143 @@ def construct( s, MsgType, num_entries=2 ):
...
+    s.full  = Wire( Bits1 )
+
+    # Logic
+
+    @s.update_on_edge
+    def up_full():
+      if s.reset:
+        s.full = b1(0)
+      else:
+        s.full = ~s.deq.en & (s.enq.en | s.full) | s.enq.en & s.full
```
jsn1993: Hmmm this seems too complicated? Does ```s.full = s.enq.en | (s.full & ~s.deq.en) work?```

pymtl3/stdlib/rtl/queues.py
jsn1993: I guess it is not that unreadable compared to the if s.reset case.

PR #54, opened 2019-06-10
Yz hypothesis demo

icyxiao96:
This PR contains Hypothesis stateful testing infrastructure using the open-loop cl pass. test_wrapper wraps an RTL component to CL using generated RTL2CL adapters. test_stateful compares a wrapped RTL with CL, which assumes cycle accuracy.


pymtl3/stdlib/test/stateful/test_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+
+    s.{name}_called = False
+    s.{name}_rdy = False
+{initialize_args}
+
+    @s.update
+    def update_{name}_adapter():
+      s.{name}_rtl.en = Bits1( 1 ) if s.{name}_called else Bits1( 0 )
+      s.{name}_called = False
+{update_args}
```
jsn1993: Add some comments saying we have to generate the adapter because setattr doesn't work for implicit value constraints. We have to compile source code to supply precise names of the arguments statically.

pymtl3/passes/GenDAGPass.py
```diff
@@ -164,6 +164,7 @@ def _process_value_constraints( self, top ):
     write_upblks = defaultdict(set)
 
     constraint_objs = defaultdict(set)
+    expl_constraints = set( top._dsl.all_U_U_constraints )
```
jsn1993: do you still need this?

icyxiao96: Got an error when it hits https://github.com/cornell-brg/pymtl3/blob/bd7da096cd0dabd4a3bbc0ddc07931a51728b323/pymtl3/passes/GenDAGPass.py#L188-L202

jsn1993: oh right. I haven't touched this part for a while ...

pymtl3/stdlib/test/stateful/test_wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+
+    s.method_specs = inspect_rtl( s.model )
+
+    # Add adapters
+    for method_name, method_spec in s.method_specs.iteritems():
+      callee_ifc = NonBlockingCalleeIfc()
+      setattr( s, method_name, callee_ifc )
+      s.connect( callee_ifc, getattr( s.model, method_name ) )
+
+    ComponentLevel6._handle_decorated_methods( s )
```
jsn1993: hmmm I don't think you need this handle_decorated_method here?

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+from pymtl3 import *
+from pymtl3.dsl.errors import InvalidConnectionError
+
+
+#-------------------------------------------------------------------------
+# CalleeRTL2CL
+#-------------------------------------------------------------------------
+class CalleeRTL2CL( Component ):
+
+  def construct( s, ifc_rtl_callee ):
```
jsn1993: I think for adapter we usually take types as arguments. See here. If you pass in ArgsType and RetType you don't need deepcopy and inverse

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+
+    ifc_rtl_caller = deepcopy( ifc_rtl_callee )
+    ifc_rtl_caller._dsl.constructed = False
+    s.ifc_rtl_caller = ifc_rtl_caller.inverse()
+
+    s.called = False
+    s.rdy = False
+
+    # generate upblk depending on args
+    if s.ifc_rtl_caller.args:
```
jsn1993: Same here. If you pass in ArgsType and RetType you can check if ArgsType is None.

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+      # has args, create tmp var
+      s.args = s.ifc_rtl_caller.args._dsl.Type()
+
+      # update args & en
+      @s.update
+      def update_en_args():
```
jsn1993: I think we discover some problems of clearing s.called after the update block. See here for our new implementation https://github.com/cornell-brg/pymtl3/blob/master/pymtl3/stdlib/ifcs/SendRecvIfc.py#L143-L155. You should be able to do something similar.

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+    return s.rets
+
+
+#-------------------------------------------------------------------------
+# CalleeIfcRTL
+#-------------------------------------------------------------------------
+class CalleeIfcRTL( Interface ):
+
+  def construct( s, ArgTypes=None, RetTypes=None ):
+    if ArgTypes:
```
jsn1993: I think we could have a CallerIfcRTL which is basically the inverse.

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+      if this.args:
+        assert other.args
+        parent.connect( other.args, this.args )
+      if this.rets:
+        assert other.rets
+        parent.connect( this.rets, other.rets )
+
+      parent.connect_pairs( this.en, other.en, this.rdy, other.rdy )
+
+    if isinstance( other, CalleeIfcRTL ):
```
jsn1993: If you want to do connect_by_name, you don't need to do anything here. Return False will revert to the default scheme.

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+      connect_callee_ifc( s, other )
+      return True
+
+    if isinstance( other, NonBlockingCalleeIfc ):
+      if s._dsl.level <= other._dsl.level:
+        raise InvalidConnectionError(
```
jsn1993: nice!

pymtl3/stdlib/ifcs/CalleeIfc.py
jsn1993: Hmm maybe we just rename the file to generic_method_ifcs.py and you implement both CalleeIfcRTL, CallerIfcRTL, and the adapter, in the same file?

pymtl3/stdlib/test/stateful/test_stateful_queues_test.py
```diff
@@ -0,0 +1,302 @@
...
+#-------------------------------------------------------------------------
+# SingleEntryPipeQueue
+#-------------------------------------------------------------------------
+class SingleEntryPipeQueue( Component ):
+
+  def construct( s, EntryType ):
+
+    # Interface
+
+    s.enq = CalleeIfcRTL( ArgTypes=[( 'msg', EntryType ) ] )
```
jsn1993: I was thinking if we should create the BitStruct here, instead of doing that in the interface, but maybe this looks better? @yo96 what do you think?

pymtl3/stdlib/ifcs/CalleeIfc.py
jsn1993: When you instantiate an adapter (e.g. https://github.com/cornell-brg/pymtl3/blob/master/pymtl3/stdlib/ifcs/SendRecvIfc.py#L49), you actually have the MsgType from when the interface is instantiated (https://github.com/cornell-brg/pymtl3/blob/master/pymtl3/stdlib/ifcs/SendRecvIfc.py#L30). Thus you can pass the interface Type into the adapter. Sometimes we even pass in two types https://github.com/cornell-brg/pymtl3/blob/master/pymtl3/stdlib/ifcs/mem_ifcs.py#L44.

pymtl3/stdlib/ifcs/CalleeIfc.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # Generate upblk and add constraints for rdy
+    @s.update
+    def update_rdy():
+      s.rdy = True if s.ifc_rtl_caller.rdy else False
+
+    s.add_constraints( U( update_rdy ) < M( cl_method ) )
+
+  @non_blocking( lambda s: s.rdy )
+  def cl_callee_method( s, **kwargs ):
```
jsn1993: I think a cleaner way to do so, is to have a separate ready method and two methods. You can instantiate an empty NonBlockingCalleeIfc() in construct, and based on the parameters, you set the calleeifc's method to either cl_callee_method, or cl_callee_method_no_args.

jsn1993: Putting non_blocking twice will create two bundles before you enter the construct

pymtl3/stdlib/test/stateful/test_stateful_queues_test.py
icyxiao96: I think this is a way to enforce that args and rets are BitStruct. Otherwise, a user can pass in anything like Bits

jsn1993: hmm maybe they could pass in a Bits object if they want? Does it change anything?

icyxiao96: the adapter is implemented based on the assumption that args is BitsStruc. And in stateful testing, it is assumed that rets is BitsStruct whe comparing results. https://github.com/cornell-brg/pymtl3/blob/c09badc088611bc13bf19aeb92fe669c5664b614/pymtl3/stdlib/ifcs/CalleeIfc.py#L85-L89

icyxiao96: Not sure if this is the best way to do it: I made CalleeIfcRTL, CallerIfcRTL and CalleeRTL2CL take ArgType and RetType directly and added two helper function to create the BitStruct: https://github.com/cornell-brg/pymtl3/blob/c5568e5735a171245d613e4a141f8fb8bceea5e0/pymtl3/stdlib/ifcs/generic_method_ifcs.py#L170-L199  In this way the user can just do https://github.com/cornell-brg/pymtl3/blob/c5568e5735a171245d613e4a141f8fb8bceea5e0/pymtl3/stdlib/test/stateful/test_stateful_queues_test.py#L33-L34 which will 1. look clean 2. Enforce that the types passed into CalleeIfcRTL are BitsStruct so that adapter and connect could work as expected

pymtl3/stdlib/ifcs/CalleeIfc.py
icyxiao96: fixed -> https://github.com/cornell-brg/pymtl3/blob/6db73d99755ce8c11ed2b97b30751cadec0f76c2/pymtl3/stdlib/ifcs/generic_method_ifcs.py#L46-L59

pymtl3/stdlib/ifcs/CalleeIfc.py
icyxiao96: fixed -> https://github.com/cornell-brg/pymtl3/blob/6db73d99755ce8c11ed2b97b30751cadec0f76c2/pymtl3/stdlib/ifcs/generic_method_ifcs.py#L145-L165

jsn1993:
@cbatten I have a new idea. How about putting Yixiao's stuff in stdlib/example/?

cbatten:
What exactly is "Yixiao's stuff"? I don't think we should have example in stdlib ... stdlib is meant to be installed ...

jsn1993:
> What exactly is "Yixiao's stuff"? I don't think we should have example in stdlib ... stdlib is meant to be installed ...

This PR has two parts. One part is the state machines and wrappers. I think these should stay in stdlib. For other things like rob test and queue test, do you think they could go to examples/?

cbatten:
yes! I think we might eventually end up with a full tutorial just on hypothesis stateful testing.

jsn1993:
Sorry guys but I think we won't merge this PR into master due to the massive effort to make it work with the latest master. I will be porting some of the important changes to a new PR. By courtesy I will be cherry-picking some of Yixiao's commits to that PR. Doing so will make sure that after the new PR is merged, Yixiao will be in the contributor list of pymtl3.

PR #56, opened 2019-06-15
Yosys Backend

ptpan:
This pull request implements the translation and import support for Yosys.


pymtl3/passes/sverilog/translation/test/SVTranslator_L1_cases_test.py
```diff
@@ -120,6 +130,7 @@ def upblk():
 
 endmodule
 """
+  a._ref_src_yosys = a._ref_src
```
jsn1993: Hmmm why are you letting this SVTranslator test be aware of yosys?

ptpan: No it's just the test case. I included the reference output of yosys backend in the same file...

ptpan: When the SV tests are run they only look at `_ref_src`. Similarly yosys tests only care about `_ref_src_yosys`

jsn1993: is it possible that you put those ref src in the yosys folder? it is kind of weird to put yosys stuff inside here ..

ptpan: Hmmm that sounds difficult because the only object returned from the test functions is a component, which does not tell you the original test function where it is constructed.

jsn1993: I guess it's a general issue when we want to simply import and reuse the whole test. I would imagine if you add another very irrelevant backend like FIRRTL, you still have to put the ref_firrtl into the sverilog test file. What do you think?

ptpan: I think a clean separation between directed tests and the reference outputs is definitely gonna help. Maybe we can centralize all directed test cases and distribute reference output in different tests. But I also kinda want to see the tests side by side with the output...

PR #57, opened 2019-06-16
Generic bits strategy

yo96:
A small PR that adds a generic bits strategy for hypothesis testing.

### DIFF
--- a/pymtl3/datatypes/strategies.py
+++ b/pymtl3/datatypes/strategies.py
@@ -19,7 +19,7 @@ from .bits_import import mk_bits
 #-------------------------------------------------------------------------
 
 @st.composite
-def bits_strat( draw, nbits, signed=False ):
+def bits( draw, nbits, signed=False ):
   if nbits == 1:
     value = draw( st.booleans() )
   elif not signed:
--- a/pymtl3/datatypes/strategies.py
+++ b/pymtl3/datatypes/strategies.py
@@ -1,0 +1,35 @@
+"""
+==========================================================================
+strategies.py
+==========================================================================
+Hypothesis strategies for built-in data types.
+
+Author : Yanghui Ou
+  Date : June 16, 2019
+"""
+from __future__ import absolute_import, division, print_function
+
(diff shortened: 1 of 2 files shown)


jsn1993:
hmm did you sync up with Peitian on this? He seems to have his own Bits strategy for the closed-loop thingy

yo96:
It's kind of different. His data strategy takes in a dut and automatically generate inputs for that component, whereas this is just a simple strategy that returns a bits object based on nbits argument.

ptpan:
My strategies are for ports and interfaces for different IR data types: `Vector`, `Struct`, and `PackedArray`. The bits strategy is similar to `Vector` so I can implement the Vector strategy using the bits strategy.

yo96:
what do you think of the naming? is `bits_strat` ok?

ptpan:
> what do you think of the naming? is `bits_strat` ok?

I kinda like `bits` over `bits_strat` because hypothesis's built-in strategies have names like `booleans`, `binary`, `decimals`. What do you think?

yo96:
from pymtl3.datatypes import strategies as st
st.bits(4).example()

yo96:
@cbatten any suggestions for naming?

cbatten:
we should ask Zac what he thinks on slack?

cbatten:
@Zac-HD what are your thoughts on how to name and organize our own bits strategy?

Zac-HD:
General advice is to pick a convention, document it for users, use it in all your own code and code samples, and stick to it.

If you want to use the same convention as Hypothesis itself - which I would suggest, since it doesn't sound like you have a firm alternative:

- Name the strategy `bits`
- Put all custom strategies with the same dependencies into a single module named `strategies` (not necessarily `datatypes.strategies`)
- Import the module, not the function, with a meaningful prefix.  For example `import hypothesis.extra.numpy as npst` or `import pymtl3.strategies as pmst`

If you have multiple distinct strategies that you'd want to call `bits`, this requires some more care.

PR #61, opened 2019-06-18
Translation and import pass improvement

ptpan:
1. Add support for translating constant struct instance
2. Add support for caching ventilated model
3. For import PyMTL wrapper, use upblk to assign a bunch of wires to a struct signal (this helps avoiding weird errors without adding explicit double buffering).


pymtl3/datatypes/BitStruct.py
```diff
@@ -24,6 +24,9 @@ class BitStruct( object ):
 
   def to_bits( self ):
     bits_lst = []
+    cls_name = self.__class__.__name__
```
jsn1993: move this line into .format() of the next line

pymtl3/datatypes/BitStruct.py
```diff
@@ -66,6 +69,9 @@ def __eq__( self, other ):
     # TODO: figure out whether we should just compare each fields
     # recursively so that it supports types that cannot be converted to bits?
     return self.to_bits() == other.to_bits()
+  
+  def __hash__( self ):
```
jsn1993: when do you need hash?

PR #63, opened 2019-06-19
Fix reset-related issues

ptpan:
1. The sequential block in import wrapper now reads in `reset` value
2. Use `RegRst` for `full` signal of bypass queue
3. Add translation support and tests for `reduce_and/or/xor`

### DIFF
--- a/pymtl3/passes/sverilog/import_/ImportPass.py
+++ b/pymtl3/passes/sverilog/import_/ImportPass.py
@@ -206,12 +206,15 @@ Fail to verilate model {} in file {}
     port_inits = '\n'.join( port_inits )
 
     # Fill in the C wrapper template
-    if not cached:
-      with open( template_name, 'r' ) as template:
-        with open( wrapper_name, 'w' ) as output:
-          c_wrapper = template.read()
-          c_wrapper = c_wrapper.format( **locals() )
-          output.write( c_wrapper )
+    # Since we may run import with or without dump_vcd enabled, we need
+    # to dump C wrapper regardless of whether the verilated model is
+    # cached or not.
+    # if not cached:
+    with open( template_name, 'r' ) as template:
+      with open( wrapper_name, 'w' ) as output:
+        c_wrapper = template.read()
+        c_wrapper = c_wrapper.format( **locals() )
+        output.write( c_wrapper )
 
     return wrapper_name, port_cdefs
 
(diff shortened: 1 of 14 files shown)


pymtl3/passes/sverilog/import_/ImportPass.py
```diff
@@ -831,7 +840,7 @@ def gen_comb_input( s, packed_ports ):
     ret = []
     for py_name, rtype in packed_ports:
       p_n_dim, p_rtype = s._get_rtype( rtype )
-      if s._get_direction( p_rtype ) == 'InPort':
+      if s._get_direction( p_rtype ) == 'InPort' and py_name != 'clk':
```
jsn1993: add some comments about why you need to do this

pymtl3/passes/sverilog/import_/ImportPass.py
```diff
@@ -223,7 +226,13 @@ def create_shared_lib( s, wrapper_name, full_name, dump_vcd, cached ):
     """Return the name of compiled shared lib."""
     lib_name = 'lib{}_v.so'.format( full_name )
 
-    if not cached:
+    # Since we may run import with or without dump_vcd enabled, we need
+    # to compile C wrapper regardless of whether the verilated model is
+    # cached or not.
+    # TODO: A better caching strategy is to attch some metadata
```
jsn1993: typo

pymtl3/passes/sverilog/import_/ImportPass.py
```diff
@@ -206,12 +206,15 @@ def create_verilator_c_wrapper( s, m, full_name, packed_ports, dump_vcd, cached
     port_inits = '\n'.join( port_inits )
 
     # Fill in the C wrapper template
```
jsn1993: It might be better to write two paragraphs here. One is what you are doing. One is what we are planning to do about the 2-level caching scheme

pymtl3/passes/sverilog/import_/verilator_wrapper.py.template
```diff
@@ -142,11 +142,12 @@ class {component_name}( Component ):
 
     @s.update_on_edge
     def seq_upblk():
+      _ffi_m.reset[0] = int(s.mangled__reset)
```
jsn1993: do you still need this?

jsn1993: hmmmm i guess this is irrelevant?

pymtl3/stdlib/rtl/enrdy_queues.py
```diff
@@ -44,7 +44,7 @@ def construct( s, Type ):
 
     s.buffer  = RegEn( Type )( in_ = s.enq.msg )
 
-    s.full = Reg( Bits1 )
+    s.full = RegRst( Bits1, reset_value = Bits1(0) )
```
jsn1993: you might only need 0 because RegRst does something like Type(reset_value). I was thinking about a register that register a BitStruct ... the current approach won't work.

PR #70, opened 2019-07-12
[WIP] Prep for release on PyPI

Zac-HD:
```
pip install --user --upgrade setuptools wheel twine
rm -r ./dist/
python setup.py sdist bdist_wheel
# Note: check version number first; upload to testpypi to confirm install works
twine upload --skip-existing dist/*
```

See [this page](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives) for details; you'll need to create an account on PyPI and I'd recommend running against testpypi first.

Finally, we'll need to change over to explicitly writing out the version number, so that installation from the source distribution (which is not a git repo) from PyPI still works.  I'd suggest copying the implementation [that I use in hypothesis-jsonschema](https://github.com/Zac-HD/hypothesis-jsonschema/blob/master/setup.py).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,30 @@ PyMTL 3 (Mamba) is the latest version of PyMTL, an open-source
 Python-based hardware generation, simulation, and verification framework with
 multi-level hardware modeling support. The original PyMTL was introduced
 at MICRO-47 in December, 2014. Please note that PyMTL 3 is currently
-**alpha** software that is under active development and documentation is
+**beta** software that is under active development and documentation is
 currently quite sparse.
+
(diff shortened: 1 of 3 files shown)


ptpan:
[Here](https://test.pypi.org/project/pymtl3) is the project page of PyMTL 3 on test-pypi. You can install PyMTL3 _without_ dependencies from test-pypi using the following command:
```
pip install --index-url https://test.pypi.org/simple/ pymtl3
```
If you have correct dependencies installed locally then you should be able to import `pymtl3` module after the above command finishes.

ptpan:
Any thoughts? @jsn1993 @yo96 @Zac-HD

Zac-HD:
Looks good to me!

PR #71, opened 2019-07-14
Implement free connect, //=, and stdlib connect_pairs

jsn1993:
This PR reimplements s.connect() as free connect(). I record a stack as I construct the model and directly query the stack top to know which component I'm in to actually call the internal connect() on the component.

I also implement //= on connectable as another handy way to do connect().

Also connect_pairs is now a stdlib thing.

I have to modify all those files that use s.connect and that's why this PR looks like a massive one.


examples/ex01_basics/IncrValueModular_test.py
```diff
@@ -38,8 +38,8 @@ def construct( s ):
     s.buf1 = Wire( Bits8 )
     s.buf2 = Wire( Bits8 )
 
-    s.connect( s.in_, s.buf1 )
-    s.connect( s.out, s.buf2 )
+    connect( s.in_, s.buf1 )
```
ptpan: So the use of shorthand `//=` versus `connect` is just arbitrary? Do we have any preference of using one over another in our own tests?

examples/ex01_basics/IncrMethodModular_test.py
```diff
@@ -60,8 +60,8 @@ def construct( s ):
     s.buf2  = Buffer()
 
     # Connect the callee ports to buf1/buf2 write/read method ports
-    s.connect( s.write, s.buf1.write )
-    s.connect( s.read,  s.buf2.read  )
+    connect( s.write, s.buf1.write )
+    connect( s.read,  s.buf2.read  )
```
ptpan: How does the system know the different between `s.connect(s.q.out, s.q.in_)` and `s.q.connect(s.q.in_, s.q.out)`?

pymtl3/dsl/ComponentLevel3.py
```diff
@@ -31,6 +40,11 @@
 from .Placeholder import Placeholder
 
 
+def connect( o1, o2 ):
+  host, o1_connectable, o2_connectable = _connect_check( o1, o2, internal=False )
+  assert host is not None, "???"
```
ptpan: Hey at least say something lol

pymtl3/dsl/Connectable.py
```diff
@@ -34,6 +35,50 @@ def get_host_component( s ):
...
+
+  if isinstance( o1, Connectable ):
+    o1_connectable = True
+    top = o1._dsl.elaborate_top
+
+  if isinstance( o2, Connectable ):
+    o2_connectable = True
+    o2_top = o2._dsl.elaborate_top
+
+    if o1_connectable: assert o2._dsl.elaborate_top is top, "???"
```
ptpan: You might want to take a pass to add some explanation for failed assertions.

pymtl3/dsl/Connectable.py
```diff
@@ -34,6 +35,50 @@ def get_host_component( s ):
       except AttributeError:
         raise NotElaboratedError()
 
+  def __ifloordiv__( s, other ):
+    host, s_connectable, o_connectable = _connect_check( s, other, internal=False )
+    assert host is not None, "???"
```
ptpan: Also here

examples/ex01_basics/IncrValueModular_test.py
jsn1993: I didn't have a good answer for this ... I arbitrarily port some s.connect to connect and some to //=.

examples/ex01_basics/IncrMethodModular_test.py
jsn1993: I implement a way for connect to know which component calls the connect. https://github.com/cornell-brg/pymtl3/pull/71/files#diff-ee19c763e84ed97152d82ad3f94b778aR184

PR #79, opened 2019-07-27
New bit struct

yo96:
In this PR I implemented new syntax for creating bit structs which uses class decorators, which is cleaner and much more powerful. It is much inspired by python3 dataclass. The syntax and mechanism are similar to dataclass too.


yo96:
@ptpan once we have finalized the syntax for our new bit struct, can you work on supporting translating the new bit struct in this branch when you have time?
My idea is that if you want to check where a class or instance is a bit struct you can do the following:
~~~python
from bit_structs import _IS_BIT_STRUCT
# _IS_BIT_STRUCT just equals to '__bit_struct_fields__'

def is_bit_struct( cls ):
  return _IS_BIT_STRUCT in cls.__dict__
~~~

To obtain the field information you can do something like this:
~~~python
def get_fields( cls ):
  fields = getattr( cls, _IS_BIT_STRUCT )
  return { name : f.type_ for name, f in fields.items() }
~~~

pymtl3/datatypes/test/bit_structs_test.py
```diff
@@ -0,0 +1,92 @@
...
+bit_structs_test.py
+==========================================================================
+Test cases for bit_structs.
+
+Author : Yanghui Ou
+  Date : July 27, 2019
+"""
+from __future__ import print_function
+
+from ..bit_structs import _IS_BIT_STRUCT, bit_struct, field, mk_bit_struct
```
jsn1993: do you need to import _IS_BIT_STRUCT here?

yo96: oh I wanted to use it to test whether a class is a bit struct but forgot to implement the test. I'll add more tests.

pymtl3/datatypes/bit_structs.py
```diff
@@ -0,0 +1,442 @@
...
+# The actual class decorator. We add a * in the argument list so that the
+# following argument can only be used as keyword arguments.
+
+def bit_struct( _cls=None, *, add_init=True, add_str=True, add_repr=True,
+                add_hash=True ):
+
+  def wrap( cls ):
+    return _process_class( cls, add_init, add_str, add_repr )
+
+  # Called as @bitstruct(...)
```
jsn1993: what does this do?

pymtl3/datatypes/bit_structs.py
```diff
@@ -0,0 +1,442 @@
...
+# - args_lst : a list of arguments in string
+# - body_lst : a list of statement of the function body in string
+#
+# Note: this function mutates _locals and should only be called internally
+# to this module.
+
+def _create_fn( fn_name, args_lst, body_lst, _globals=None, _locals=None ):
+
+  # Lazily construct empty dictionary
+  if _locals is None:
```
jsn1993: what about _globals?

pymtl3/datatypes/bit_structs.py
yo96: This is to check if the decorator has any argument. We can use the decorator like
~~~
@bit_struct
class A:
  ...
~~~
or 
~~~
@bit_struct( add_repr = False )
class A:
  ...
~~~

pymtl3/datatypes/bit_structs.py
yo96: Oh I think I did this because as I tried `exec` can take globals as none but not locals.

yo96: if locals is None then exec will mutate the current locals()

jsn1993: ok!

jsn1993: maybe add some comments on this exec thing here

pymtl3/datatypes/bit_structs.py
jsn1993: Hmm this new argument seperator might be able to solve some problem I encountered when I was playing around with method port decorator.

pymtl3/datatypes/bit_structs.py
```diff
@@ -0,0 +1,442 @@
...
+    [ _mk_init_body( self_name, f ) for f in fields ],
+    _locals  = _locals,
+    _globals = _globals,
+  )
+
+#-------------------------------------------------------------------------
+# _mk_str_fn
+#-------------------------------------------------------------------------
+# Creates a __str__ function based on fields.
+
```
jsn1993: Since you are generating some code from a list here, can you provide one example for each function on the actual code it generates?

pymtl3/datatypes/bit_structs.py
```diff
@@ -0,0 +1,476 @@
...
+  cls_fields = [ _get_field( cls, a_name, a_type )
+                 for a_name, a_type in cls_annotations.items() ]
+
+  # Create a dictionary of fields
+  for f in cls_fields:
+    fields[ f.name ] = f
+
+  # Stamp the special attribute so that translation pass can identify it
+  # as bit struct.
+  setattr( cls, _IS_BIT_STRUCT, fields )
```
jsn1993: No need to use setattr here because the attribute is a constant string

pymtl3/datatypes/bit_structs.py
```diff
@@ -0,0 +1,476 @@
...
+#-------------------------------------------------------------------------
+# Creates a __eq__ function based on fields. By default it just compares
+# each field. For example, if fields contains two field x (Bits4) and y
+# (Bits4), _mk_init_fn will return a function that looks like the
+# following:
+#
+# def __eq__( self, other ):
+#   if other.__class__ is self.__class__:
+#     return (self.x,self.y,) == (other.x,other.y,)
+#   else:
```
jsn1993: I wonder if every struct comparison needs to call  .__class__. If so, it might hurt performance. Maybe you could use try-catch block to do this, something like:
```
try:
     return (self.x,self.y,) == (other.x,other.y,)
except Exception as e:
    <handle the exception -- print struct comparison failure>
```

pymtl3/datatypes/bit_structs.py
yo96: I keep it this way so that we can easily change this constant string. This won't hurt performance right?

pymtl3/datatypes/bit_structs.py
yo96: hmm that way we are allowing comparison between different types? Than may cause Pymtl behaves differently from Verilator though

jsn1993: On my end, I think the if statement and comparisons involve __class__ will kill the JIT...

pymtl3/datatypes/bit_structs.py
jsn1993: Since it's not during runtime, it shouldn't hurt.

pymtl3/datatypes/bit_structs.py
jsn1993: @ptpan @yodada nominal vs structural typing gurus please chime in here

yo96: i don't think we ever directly compare two bit structs except in the test harness... i think the default `__eq__` should be conservative and should not behave differently from that translated verilog. Besides, user can just overwrites the default `__eq__` if they are not happy with it

jsn1993:
Feature request -- can we add a copy method similar to deepcopy but is called on a bit struct object, and it recursively creates a new struct with the results of calling copy to the child fields. i need to implement it Bits as well

jsn1993:
I will be working on this to get it merged ASAP

yo96:
why making `mk_msg` a class method? I don't think it should be a first class thing. We can always just pass in bits objects to `__init__`. If user wants type cast they can just write a helper function that does that.

yo96:
I don't think it works for nested struct either

jsn1993:
> I don't think it works for nested struct either

nested struct will work recursively like NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) ), in which case the user has to create helper methods for both NetMsg and MemReqMsg.

jsn1993:
> why making `mk_msg` a class method? I don't think it should be a first class thing. We can always just pass in bits objects to `__init__`. If user wants type cast they can just write a helper function that does that.

I just want to release the burden from the users because they might have to create helper method in every test file in the code base that creates a message type.

yo96:
> > I don't think it works for nested struct either
> 
> nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.

They can just create one helper function that flattens all the fields.I think we should keep things simple and give user more freedom.

yo96:
> > I don't think it works for nested struct either
> 
> nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.

Also does this actually work? `MemReqMsg.mk_msg(...)` returns an object but can the  `__init__` method of `MemReqMsg` takes in an object?

jsn1993:
> > > I don't think it works for nested struct either
> > 
> > 
> > nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.
> 
> They can just create one helper function that flattens all the fields.I think we should keep things simple and give user more freedom.

what do you mean by flattening all the fields?

jsn1993:
> > > I don't think it works for nested struct either
> > 
> > 
> > nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.
> 
> Also does this actually work? `MemReqMsg.mk_msg(...)` returns an object but can the `__init__` method of `MemReqMsg` takes in an object?
```
@classmethod
def mk_msg(cls, x, y):
  return cls(Bits4(x), Bits8(y))
``` this is how it works with only Bits.
For BitStruct as a subfield, I just assign it like __init__
```@classmethod
def mk_msg(cls, pt0, pt1):
  return cls(pt0, pt1)
```

jsn1993:
@ptpan i already fixed all the translation related issues and make all the tests pass under cpy, pypy, pypy+mamba.

yo96:
> > > > I don't think it works for nested struct either
> > > 
> > > 
> > > nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.
> > 
> > 
> > They can just create one helper function that flattens all the fields.I think we should keep things simple and give user more freedom.
> 
> what do you mean by flattening all the fields? Also, adding such method won't give the user less freedom. They can also override the mk_msg method if they want.

If I have a nested struct that has a point field and a pixel field I can just write a helper function `mk_ptpxl_msg( x, y, r, g, b )` rather than calling the mk_msg for both class.

In our update block we basically leave it for the user to handle the types. Why can't we do this for bit struct

jsn1993:
We don’t force them to use our mk_msg. Like you said we can use mk_ptpxl_msg and totally ignore mk_msg. The existence of mk_msg doesn’t conflict with what you said. It’s just some helper method that the user could leverage without creating their own stuff.

yo96:
I don't think this is very useful and I don't quite like the syntax. Plus it adds one more thing for us to maintain.

jsn1993:
> > > > > I don't think it works for nested struct either
> > > > 
> > > > 
> > > > nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.
> > > 
> > > 
> > > They can just create one helper function that flattens all the fields.I think we should keep things simple and give user more freedom.
> > 
> > 
> > what do you mean by flattening all the fields? Also, adding such method won't give the user less freedom. They can also override the mk_msg method if they want.
> 
> If I have a nested struct that has a point field and a pixel field I can just write a helper function `mk_ptpxl_msg( x, y, r, g, b )` rather than calling the mk_msg for both class.
> 
> In our update block we basically leave it for the user to handle the types. Why can't we do this for bit struct

In your NetMsg example, creating a flat function may not be the desired thing to do because the MemMsg type might NOT be visible to you. It can be parameterized and provide to you as mk_net_msg( ..., payloadType ). How do you create a flat function without inspecting the fields of payloadType?

> I don't think this is very useful and I don't quite like the syntax. Plus it adds one more thing for us to maintain.

For the syntax, we have this in v2 https://github.com/cornell-brg/pymtl/blob/master/pclib/ifcs/MemMsg.py#L80-L102 and we found it quite useful because we can just forget about the bitwidth when we create a message from the beginning. Like here https://github.com/cornell-brg/pymtl3/blob/yo-dataclass-bitstruct/pymtl3/stdlib/ifcs/MemMsg_test.py#L112. @cbatten can chime in here and comment on whether v3 should provide something similar.

In your example "In our update block we basically leave it for the user to handle the types.", leaving it for the user makes the user do less work and hence improve their productivity instead of getting everything iron solid at early phase of development. In our case of providing a classmethod to the user to create messages without thinking about width and the helper function, so the user does less work and hence becomes more productive. In my opinion, it doesn't hurt to add more non-unuseful things in the framework as long as it doesn't conflict with everything else.

Lastly, how much additional effort do you think we need to maintain this when it's basically the same thing as ```__init__```?

jsn1993:
> > > > > > I don't think it works for nested struct either
> > > > > 
> > > > > 
> > > > > nested struct will work recursively like `NetMsg.mk_msg( src_x, src_y, MemReqMsg.mk_msg(...) )`, in which case the user has to create helper methods for both NetMsg and MemReqMsg.
> > > > 
> > > > 
> > > > They can just create one helper function that flattens all the fields.I think we should keep things simple and give user more freedom.
> > > 
> > > 
> > > what do you mean by flattening all the fields? Also, adding such method won't give the user less freedom. They can also override the mk_msg method if they want.
> > 
> > 
> > If I have a nested struct that has a point field and a pixel field I can just write a helper function `mk_ptpxl_msg( x, y, r, g, b )` rather than calling the mk_msg for both class.
> > In our update block we basically leave it for the user to handle the types. Why can't we do this for bit struct
> 
> In the NetMsg example, creating a flat function may not be the desired thing to do because the MemMsg type might NOT be visible to you. It can be parameterized and provide to you as mk_net_msg( ..., payloadType ). How do you create a flat function without inspecting the fields of payloadType?
> 
> > I don't think this is very useful and I don't quite like the syntax. Plus it adds one more thing for us to maintain.
> 
> For the syntax, we have this in v2 https://github.com/cornell-brg/pymtl/blob/master/pclib/ifcs/MemMsg.py#L80-L102 and we found it quite useful because we can just forget about the bitwidth when we create a message from the beginning. Like here https://github.com/cornell-brg/pymtl3/blob/yo-dataclass-bitstruct/pymtl3/stdlib/ifcs/MemMsg_test.py#L112. @cbatten can chime in here and comment on whether v3 should provide something similar.
> 
> In your example "In our update block we basically leave it for the user to handle the types.", leaving it for the user makes the user do less work and hence improve their productivity instead of getting everything iron solid at early phase of development. In our case of providing a classmethod to the user to create messages without thinking about width and the helper function, so the user does less work and hence becomes more productive. In my opinion, it doesn't hurt to add more non-unuseful things in the framework as long as it doesn't conflict with everything else.
> 
> Lastly, how much additional effort do you think we need to maintain this when it's basically the same thing as `__init__`?

cbatten:
I think we should start by not having mk_msg. We should use the constructor. The reason we had mk_msg is because the v2 bit structs didn’t have constructors!

jsn1993:
It’s different again. V2 doesn’t worry about type casting.

jsn1993:
To be clear here, the init takes the argument and copies it in as is. This doesn’t do any type casting so if you pass in an int it stores it as int. The goal of mk_msg is to provide another constructor that does cast those int to bits.

pymtl3/datatypes/bit_struct.py
```diff
@@ -0,0 +1,535 @@
...
+  _globals = {}
+
+  assign_strs = []
+  for f in fields:
+    type_ = f.type_
+    if isinstance( type_, list ) or is_bit_struct( type_ ):
+      assign_strs.append( f'{f.name}' )
+    else:
+      assert issubclass( type_, Bits )
+      _globals[ type_.__name__ ] = type_
```
yo96: i don't think using type.__name__ is a good idea. If we have two types that has the same name but from different modules this wouldn't work. For example I want to use a `a.Point` and `b.Point` in my nested struct this wouldn't work. That why I rename all types to `_type_fname` in the dictionary but you seemed to have changed it

yo96:
Well if you insist. I don't think not doing type cast is a big deal. We are doing that basically everywhere why make bit struct a special case.

jsn1993:
Good point. I will add it back.

jsn1993:
> Well if you insist. I don't think not doing type cast is a big deal. We are doing that basically everywhere why make bit struct a special case.

Again we do it everywhere doesn't mean the framework doesn't need to provide the ability to make the user more productive.

pymtl3/datatypes/bit_struct.py
jsn1993: i will add it back

jsn1993: well if you had a test to catch this I won't remove it

yo96:
I'd prefer keeping things clean, simple, and consistent. If we sometimes use plain integers and sometimes use bits objects that may confuse the user.

pymtl3/passes/rtlir/rtype/RTLIRType.py
```diff
@@ -556,7 +556,7 @@ def _is_of_type( obj, Type ):
   ( dsl.InPort,               _handle_InPort ),
   ( dsl.OutPort,              _handle_OutPort ),
   ( dsl.Wire,                 _handle_Wire ),
-  ( ( int, Bits, BitStruct ), _handle_Const ),
+  ( ( int, Bits ),          _handle_Const ),
```
ptpan: Indentation

pymtl3/passes/yosys/translation/test/YosysTranslator_L2_cases_test.py
```diff
@@ -46,4 +46,5 @@ def local_do_test( m ):
   m.elaborate()
   tr = YosysTranslator( m )
   tr.translate( m )
+  print(tr.hierarchy.src)
```
ptpan: Remove this line?

pymtl3/datatypes/bitstructs.py
```diff
@@ -0,0 +1,464 @@
...
+  # default is always None
+  if isinstance( type_, list ) or is_bitstruct( type_ ):
+    return f'{name} = None'
+  return f'{name} = 0'
+
+#-------------------------------------------------------------------------
+# _mk_init_body
+#-------------------------------------------------------------------------
+# Creates one line of __init__ body from a field and add its default value
+# to globals.
```
yo96: maybe clean up this comment? we no longer do default values

yo96:
found some leftover comments. I'll take a pass to clean up.

Issue #78, opened 2019-07-25
Integration with other Verilog-based projects


ptpan:
To better support other Verilog/SystemVerilog-based projects that use Verilator for simulation, it would be nice to have the following features:
1. Passing customized file list to Verilator (`-f` option)
2. Support for more Verilator options, especially those used to suppress warnings
3. Linking the shared library against other libraries
4. #73 
5. Instantiating a Verilog module with parameters
6. More advanced flags such as setting `rpath` during the actual C++ test bench compilation

ptpan:
@yo96 Any other things you would like to have for black-parrot testing?

yo96:
These look good to me. I think maybe we can organize all the flags read by import pass into an object called `settings` (like hypothesis does) so that we can set flags like this:
~~~
s.import_settings = ImportSettings(
  import = True,
  src_path = '../queue.v',
  list_path = '../flist.verilator',
)
~~~
This way it may be easier to manage all these flags (I'm sure there will be a ton of flags as we make the import pass more flexible) and it also saves name space. Anyway the general goal is to make the import as seamless as possible!

ptpan:
@yo96 Can you please remind me of why we wanted a special parameter `import_params` to the construction method of some placeholder to be imported?

yo96:
To compute the type for some ports

ptpan:
Closed as #80 has been merged into master

PR #80, opened 2019-07-27
Import pass enhancement

ptpan:
This pull request addresses issues mentioned in #78. To enable generic pass configuration I implemented `PassConfig` base class which provides a clean way to add, check, and use options to a pass. `ImportPass` of SystemVerilog backend extends this base class to implement its own options.


ptpan:
This pull request adds the following features to the import pass
1. An import pass configuration class that extends a generic pass configuration base class. This provides a clean approach to set up options with default values, check option values, and interface with the pass that requires these options. 
The import pass configuration class currently supports options that specify
 - Verilator source file, file list, include directory, and output make directory
 - Verilator 
2.

pymtl3/passes/PassConfigs.py
```diff
@@ -0,0 +1,63 @@
...
+
+  def get_option( s, opt ):
+    """Return the value of option `opt`."""
+    return s.options[opt]
+
+  def set_option( s, opt, value ):
+    s.options[opt] = value
+
+  def _gen_options( s, _opts ):
+    opts = copy( s.Options )
```
jsn1993: Is this s.Options still a valid variable after you comment it out?

pymtl3/passes/sverilog/import_/ImportConfigs.py
```diff
@@ -0,0 +1,452 @@
...
+      # We enforce the GNU makefile implicit rule that `LDFLAGS` should only
+      # include non-library linker flags such as `-L`.
+      "ld_flags" : "",
+
+      # We enforce the GNU makefile implicit rule that `LDLIBS` should only
+      # include library linker flags/names such as `-lfoo`.
+      "ld_libs" : "",
+    }
+
+    s.PassName = 'sverilog::ImportPass'
```
jsn1993: Will Python style instead of c++ be better here?

pymtl3/passes/sverilog/import_/ImportPass.py
```diff
@@ -105,100 +119,108 @@ def get_imported_object( s, m ):
...
     Create a C wrapper that calls verilator C API and provides interfaces
     that can be later called through CFFI.
     """
-    component_name = full_name
-    try:
-      vcd_timescale = m.vcd_timescale
-    except AttributeError:
-      vcd_timescale = "10ps"
+    p_map = config.get_port_map() if config.is_port_mapped() else lambda x: x
+    component_name = config.get_top_module()
```
jsn1993: Did you implement the 2-level caching we discussed in this PR?

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
```diff
@@ -34,7 +34,7 @@ def run_sim( _th ):
   try:
     _th.elaborate()
     _th.q.yosys_translate = True
-    _th.q.yosys_import = True
+    _th.q.yosys_import = ImportConfigs()
```
jsn1993: What is the purpose of instantiating ImportConfigs() here?

pymtl3/passes/PassConfigs.py
ptpan: None of the methods of the base class is supposed to be called directly. `s.Options` is an attribute that derived classes are supposed to override. I think making `Options` an abstract property might help this by raising an exception every time an abstract `Options` is used.

pymtl3/passes/sverilog/import_/ImportConfigs.py
ptpan: I agree

pymtl3/passes/sverilog/import_/ImportPass.py
ptpan: No not in this PR.

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
ptpan: Most of the import configurations implemented in this PR should go with the component to be imported. That's why I find it natural to reuse the `sverilog_import` attribute from which the import pass extracts configurations.

pymtl3/passes/sverilog/import_/test/ImportedObject_test.py
```diff
@@ -76,3 +82,47 @@ def construct( s ):
...
+  class VQueue( Placeholder, Component ):
+    def construct( s, data_width, num_entries, count_width ):
+      s.count   =  OutPort( mk_bits( count_width )  )
+      s.deq_en  =  InPort( Bits1  )
+      s.deq_rdy = OutPort( Bits1  )
+      s.deq_msg = OutPort( mk_bits( data_width ) )
+      s.enq_en  =  InPort( Bits1  )
+      s.enq_rdy = OutPort( Bits1  )
+      s.enq_msg =  InPort( mk_bits( data_width ) )
+      s.sverilog_import = ImportConfigs(vl_src = get_dir(__file__)+'VQueue.sv')
```
jsn1993: if it's the same dir, can the user uses 'VQueue.sv' by default?

ptpan: Most of the cases we would like to run this in a build directory and we seldom put source files under a build directory. But if somebody does this, they can use the filename alone.

pymtl3/passes/PassConfigs.py
```diff
@@ -0,0 +1,73 @@
...
+  """Base class of customized pass configrations.
+
+  Users who wish to use this Base class should (1) put the name of their
+  pass to `PassName`; (2) fill their own options and default values into
+  `Options` class attribute; (3) define `check_<option_name>` methods
+  that takes the value of `<option_name>` and returns whether this
+  option is valid.
+  """
+  def __init__( s, **kwargs ):
+    # s.Options = {}
```
jsn1993: I guess you can remove these then

pymtl3/passes/sverilog/import_/test/ImportedObject_test.py
jsn1993: good point. Running this in build directory is very likely.

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
jsn1993: My question was, since you probably need to specify import configs in the Placeholder, why do you still need to do this at the top level? Are you doing this because translation pass doesn't handle import?

pymtl3/passes/sverilog/import_/verilator_wrapper.py.template
```diff
@@ -140,16 +144,21 @@ class {component_name}( Component ):
...
-    def seq_upblk():
-      # Advance the clock
-      _ffi_m.clk[0] = 0
-      _ffi_inst.eval( _ffi_m )
-      _ffi_m.clk[0] = 1
-      _ffi_inst.eval( _ffi_m )
-      set_output()
-
-    {constraint_str}
+    if {has_clk}:
```
jsn1993: most of the time, has_clk should be true .. right?

ptpan: Yes

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
ptpan: I think sometimes you still need to specify configs at top level. It's true that most Placeholders to be imported should have such configs by default, but these DUT's actually come from `stdlib` which certainly should not come with preset configs.

jsn1993: Let me clarify. I think in this case stdlib component is a valid pymtl component with valid logic. Only when you translate it using yosys backend do you need to import it back. I was asking whether translation pass has a default import config to help something but seems like translation pass is what it is and we need to supply another import config before we apply import pass. In the case of translate-import we actually know the file names and such, maybe after the translate pass you can add a default import config to the component? Or we have a high level translateimport pass that wraps what you do here.

ptpan: Ah I see. The default values of `ImportConfigs` are set to be used with translation-import scenarios. To perform a translation-import you simply use `ImportConfigs()` to mark some component as to be imported instead of writing up all the options.

jsn1993: also i just noticed that there is a typo in the file name

jsn1993: Maybe the translation pass adds a default import configs to the component?

ptpan: That sounds weird to me... I think ideally the translation pass does what it does and we provide default import configs that work for the most common cases (translation import).

jsn1993: I mean if the user only uses translation-import in the most basic case we probably shouldn't bother them with import config.

pymtl3/passes/sverilog/import_/verilator_wrapper.c.template
```diff
@@ -163,3 +164,14 @@ void eval( V{component_name}_t * m ) {{
   #endif
 
 }}
+
+//------------------------------------------------------------------------
+// assert_en()
+//------------------------------------------------------------------------
+// Enable or disable assertions controlled by --assert
+
```
jsn1993: if i remember correctly, the assertion error in C land will turn into segmentation fault that python land cannot catch and directly terminate the simulation.

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
jsn1993: What do you think? I was basically thinking that we may add the default import config to the component in the translation pass so that the user can directly apply import pass.

pymtl3/passes/sverilog/util/utility.py
```diff
@@ -25,3 +30,13 @@ def get_string( obj ):
     assert arg_name != ''
     comp_name += '__' + arg_name + '_' + get_string(arg_value)
   return comp_name
+
+def wrap( s ):
+  col = shutil.get_terminal_size().columns
+  return "\n".join(sum((textwrap.wrap(line, col) for line in s.split("\n")), []))
+
+def expand(v):
```
jsn1993: ```( v )```

pymtl3/passes/sverilog/util/utility.py
```diff
@@ -25,3 +30,13 @@ def get_string( obj ):
...
+  return os.path.expanduser(os.path.expandvars(v))
+
+def get_dir(cur_file):
```
jsn1993: same here

pymtl3/passes/yosys/test/TranslationImport_closed_loop_dircted_test.py
ptpan: I think it should be a translation-import pass. I'll add that.

pymtl3/passes/sverilog/import_/verilator_wrapper.c.template
ptpan: It's possible to overwrite Verilator's `stop` function and swap in some code that calls Python C API to raise a Python exception. We can find out later.

jsn1993: hmmm not sure if Python C API is restricted to only CPython. Anyways, maybe you could add some TODO to record  this conversation.

pymtl3/passes/sverilog/import_/verilator_wrapper.py.template
```diff
@@ -157,6 +157,14 @@ class {component_name}( Component ):
       {constraint_str}
 
   def assert_en( s, en ):
+    # TODO: for verilator, any assertion failure will cause the C simulator
+    # to abort, which results in a Python internal error. A better approach
+    # is to throw a Python exception at the time of assertion failure.
```
jsn1993: great1

PR #81, opened 2019-08-02
Test sink enhancement

yo96:
In this PR I added the following feature:
- raises exception at the start of next cycle so that the errored line trace gets printed out
- supports customized compare function

### DIFF
--- a/pymtl3/stdlib/test/src_sink_test.py
+++ b/pymtl3/stdlib/test/src_sink_test.py
@@ -212,12 +212,9 @@ def test_error_more_msg():
       sink_msgs = [ b16(0xface) ],
     )
     th.run_sim()
-    errored = False
   except Exception as e:
-    errored = True
-    print( e )
-
-  assert errored
+    return
+  raise Exception( 'Failed to detect error!' )
 
 def test_error_wrong_msg():
   try:
@@ -227,12 +224,9 @@ def test_error_wrong_msg():
       sink_msgs = [ b16(0xface), b16(0xdead) ],
     )
     th.run_sim()
-    errored = False
   except Exception as e:
-    print( e )
-    errored = True
-
-  assert errored
+    return
+  raise Exception( 'Fail to detect error!' )
 
 def test_error_late_msg():
   try:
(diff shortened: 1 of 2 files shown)


yo96:
@jsn1993 can we merge this in real quick? It's just some minor changes but it will be very helpful for debugging RTL as I can see what's wrong in the line trace.

pymtl3/stdlib/test/src_sink_test.py
```diff
@@ -199,3 +199,67 @@ def test_adaptive( src_level, sink_level, msgs, src_init,  src_intv,
...
+
+def test_error_more_msg():
+  try:
+    th = TestHarnessSimple(
+      Bits16, TestSrcCL, TestSinkCL,
+      src_msgs  = [ b16(0xface), b16(0xface) ],
+      sink_msgs = [ b16(0xface) ],
+    )
+    th.run_sim()
+    errored = False
```
jsn1993: You don't need this variable. Just return in the except block and raise another exception instead of assert errored.

pymtl3/stdlib/test/src_sink_test.py
```diff
@@ -199,3 +199,67 @@ def test_adaptive( src_level, sink_level, msgs, src_init,  src_intv,
...
+      Bits16, TestSrcCL, TestSinkCL,
+      src_msgs  = [ b16(0xface), b16(0xface) ],
+      sink_msgs = [ b16(0xface), b16(0xdead) ],
+    )
+    th.set_param( 'top.src.construct', initial_delay=5 )
+    th.set_param( 'top.sink.construct', arrival_time=[1,2] )
+    th.run_sim()
+    errored = False
+  except Exception as e:
+    print( e )
```
jsn1993: take a pass to clean this up

pymtl3/stdlib/test/test_sinks.py
```diff
@@ -42,6 +46,19 @@ def construct( s, Type, msgs, initial_delay=0, interval_delay=0,
 
     @s.update
     def up_sink_count():
+      # Raise exception at the start of next cycle so that the errored
+      # line trace gets printed out
+      if s.error_msg:
```
jsn1993: You might have to put these blocks into an explicit non-reset phase.

PR #84, opened 2019-09-05
VCD generation pass improvement

ptpan:
This pull request plans to address the following issues with `VcdGenerationPass`:
1. It segfaults CPython while applied to a mid-sized network simulation of ~23000 signals. The segfault happens when trying to dynamically compile the `dump_vcd` function which has about two hundred thousand lines of code.
2. It raises an exception while trying to dump a BitStruct signal.


ptpan:
The implementation in this PR currently only dumps the entire BitStruct signals (i.e. it does not dump the fields as separate signals).

pymtl3/passes/VcdGenerationPass.py
```diff
@@ -81,6 +79,57 @@ def _gen_vcd_symbol():
...
 
+    # The actual dump_vcd function to be appended to _sched.
+    # The framework will call this function without arguments, and therefore
+    # I chose to monkey patch the data structure objects to the global
+    # namespace of `dump_vcd`. This might be hacky but probably better than
+    # compiling a million lines of code to a Python function (you will almost
+    # certainly get a segfault from that).
+    # TODO: make this less hacky? type check?
+
+    def dump_vcd():
```
jsn1993: You already have access to top here. I guess a less hacky way is to write a function that return a function without argument.
```
def vcd_gen(top,vcdfile):
  def dump_vcd():
    # use top/vcdfile
```

pymtl3/passes/VcdGenerationPass.py
```diff
@@ -3,16 +3,14 @@
 VcdGenerationPass.py
 ========================================================================
 
-Author : Shunning Jiang
+Author : Shunning Jiang, Yanghui Ou, Peitian Pan
```
jsn1993: You might want to change the date too.

pymtl3/passes/VcdGenerationPass.py
```diff
@@ -81,6 +79,57 @@ def _gen_vcd_symbol():
...
+
+    def dump_vcd():
+      s             = top
+      vcd_file      = vcdmeta.vcd_file
+      clock_net_idx = vcdmeta.clock_net_idx
+      clock_symbol  = net_symbol_mapping[vcdmeta.clock_net_idx]
+      next_neg_edge = 100*vcdmeta.sim_ncycles+50
+      next_pos_edge = 100*vcdmeta.sim_ncycles+100
+
+      try:
```
jsn1993: As far as I can see, you are turning the dynamic compilation for each signal into a loop that iterates over all the provided signals. I think the loop approach is better than code generation, since the "simulation performance improvement" that dynamic compilation is supposed to provide is unclear here -- we might want to see if dumping vcd is bottlenecked by disk (which I think so).

pymtl3/passes/VcdGenerationPass.py
```diff
@@ -81,6 +79,57 @@ def _gen_vcd_symbol():
...
+        # Dump VCD
+        for i, _net in enumerate( trimmed_value_nets ):
```
jsn1993: I think if you do the previous two nested function approach, the eval expression over trimmed_value_nets can be pre-processed in the outer function. Also, thanks for adding this type check thing. I think you can also store last_value in a list or something instead of always doing getattr.

pymtl3/passes/VcdGenerationPass.py
```diff
@@ -130,7 +185,8 @@ def _gen_vcd_symbol():
 
     # Vcd file takes a(0) instead of a[0]
     def vcd_mangle_name( name ):
-      return name.replace('[','(').replace(']',')')
+      # signal names with colons in it silently fail gtkwave
+      return name.replace('[','(').replace(']',')').replace(':', '_$_')
```
jsn1993: hmmm why would a signal have colon in its name?

pymtl3/passes/VcdGenerationPass.py
ptpan: Yeah good point. I think that's actually the pythonic way how people add some variables to the closure of a function...

pymtl3/passes/VcdGenerationPass.py
ptpan: A sliced signal is a leaf signal and has a name like `s.x[0:3]`. This happened when we tried to dump VCD for a mesh network.

pymtl3/passes/VcdGenerationPass.py
ptpan: What will be the performance benefit of moving the `eval` to the outside function?

jsn1993: You are correct. The reason why I needed to generate those statements is to eval the names ..

jsn1993: I guess the eval thing can be very slow if it needs to be called every cycle. Did you guys try the 23000 signal vcd again?

pymtl3/passes/VcdGenerationPass.py
jsn1993: Oh, I think we don't need to dump those signals in the vcd because the non-leaf s.x will be dumped.

ptpan: Well I think the point is sometimes you do want to see a slice of some signal because that's interesting part of a concatenated vector. Plus I think this leaf slice only exists if it appears in the design. So keeping that might actually be helpful..

pymtl3/passes/VcdGenerationPass.py
ptpan: I remember generating vcd for a large network is observably slower than running just the simulation. There is a lag before the simulation starts (probably recursing the hierarchy and finishing the initialization) and then every cycle takes longer time.

Issue #86, opened 2019-09-07
The use of `$` as separators during translation upsets some ASIC tools


ptpan:
We currently use `$` as a separator in the translation passes. Despite the fact that DC can handle signal names with `$`s pretty well, some ASIC tools fail subtly if they see signal names with `$`. For example, Synopsys vcat utility rejects a configuration file that mentions dollar-sign signals in the VCD file, without mentioning the use of `$`.

If `$` are generally not used in common Verilog source code, then we probably should consider switch to other characters (maybe `_`?) for the sake of better integration with other tools... Thoughts? @jsn1993

cbatten:
This is a great point. They key is to also avoid generating names that might conflict with names chosen by the designer. Maybe a double underscore since it is rare for a designer to use __?

ptpan:
Closing this issue because #87 has been merged

Issue #75, opened 2019-07-19
Better names for translated components with parameters


ptpan:
The current PyMTL translation pass generates a pretty long name that includes the names and values of parameters. This naming scheme becomes awkward when a component has several parameters and when these parameters have long names. We even experienced verilator compilation error when the translated name of some component grows too long (> 100 characters?). Also there are situations where customized component names are useful (e.g. integrating PyMTL generated designs with larger projects where we want to swap in PyMTL generated components without changing the external design code; also in our ASIC flow scripts it would be nice to use the same name for the top component).

We need (1) a way to make component names shorter (hashing?) and (2) to provide support for customized component names (a mapping between parameters and name?).

jsn1993:
Can we close this since #93 is already merged?

ptpan:
> Can we close this since #93 is already merged? @ptpan

The (2) part is still not done yet

jsn1993:
@ptpan is this done by our new placeholder/translate/import pass?

PR #87, opened 2019-09-09
Deprecate dollar sign during translation

ptpan:
This pull request addresses #86 by replacing the occurrences of dollar signs with double underscores.


jsn1993:
The signal name with N levels of hierarchy will be 5 characters longer than before if we use double underscore. Is that a problem? I guess it's not a big problem?

ptpan:
> The signal name with N levels of hierarchy will be 5 characters longer than before if we use double underscore. Is that a problem? I guess it's not a big problem?

This in most cases is not a problem. However if it’s the component name, Verilator might fail due to excessively long names as #75 indicates, and I plan to address that in another pull request.

PR #90, opened 2019-09-16
Implement non-recursive RPO and add a deep dag test

jsn1993:
This PR implements non-recursive RPO algorithm to avoid hitting CPython recursion depth limit.


jsn1993:
This PR solves the recursion limit problem. @yo96

PR #96, opened 2019-09-26
Change critical formatted string to fstring

jsn1993:
Since we have moved to python3, I think we'd mostly like to use fstring. Seems like fstring is slightly faster than .format. I don't think we need to change those .format or %s in those exceptions, so I only changed those critical usages.


yo96:
Looks good! But should we use single quote `'` instead of double quote `"` in the future? Since you can put `" "` inside `' '` but not the other way around.

pymtl3/dsl/Connectable.py
```diff
@@ -117,7 +117,7 @@ def __init__( s, Type, v, parent ):
     s._dsl.parent_obj = parent
 
   def __repr__( s ):
-    return "{}({})".format( str(s._dsl.Type.__name__), s._dsl.const )
+    return f"{str(s._dsl.Type.__name__)}({s._dsl.const})"
```
yo96: you probably don't need to explicitly call `str` here

pymtl3/dsl/test/sim_utils.py
```diff
@@ -65,21 +65,21 @@ def simple_sim_pass( s, seed=0xdeadbeef ):
         readers = [ x for x in signals if x is not writer ]
 
         fanout  = len( readers )
-        wstr    = repr(writer)
-        rstrs   = [ repr(x) for x in readers ]
 
-        upblk_name = "{}__{}".format(repr(writer), fanout)\
+        upblk_name = f"{repr(writer)}__{fanout}" \
```
yo96: f'{writer!r}'

pymtl3/dsl/test/sim_utils.py
```diff
@@ -65,21 +65,21 @@ def simple_sim_pass( s, seed=0xdeadbeef ):
...
                         .replace( "(", "_" ).replace( ")", "" )
 
-        src = """
-        def {0}():
-          common_writer = {1}
-          {2}
-        _recent_blk = {0}
-        """.format( upblk_name, wstr, "; ".join(
-                    [ "{} = common_writer".format( x ) for x in rstrs ] ) )
+        rstrs   = [ f"{repr(x)} = _w" for x in readers ]
```
yo96: same thing here. you can use `f'{x!r} = _w'` to make it call `repr` for you

pymtl3/dsl/test/sim_utils.py
```diff
@@ -65,21 +65,21 @@ def simple_sim_pass( s, seed=0xdeadbeef ):
...
+
+        src = f"""
+        def {upblk_name}():
+          _w = {repr(writer)}
```
yo96: `{writer!r}`

pymtl3/dsl/test/sim_utils.py
jsn1993: interesting, didn't know this

jsn1993:
> Looks good! But should we use single quote `'` instead of double quote `"` in the future? Since you can put `" "` inside `' '` but not the other way around.

I think you can put double quote inside single quote, it's just that you cannot put the same quote inside itself.
```
>>> print('"12"')
"12"
>>> print("'12'")
'12'
>>> 
```

yo96:
> > Looks good! But should we use single quote `'` instead of double quote `"` in the future? Since you can put `" "` inside `' '` but not the other way around.
> 
> I think you can put double quote inside single quote, it's just that you cannot put the same quote inside itself.
> 
> ```
> >>> print('"12"')
> "12"
> >>> print("'12'")
> '12'
> >>> 
> ```

oh yeah I think you are right. they are the same.

jsn1993:
@yo96 I realized I didn't clean up the whole folder. You may have to review a LOT of files. also @ptpan you should also take a look since I cleaned up all those RTLIR and translation passes and just want to fill you in about what I did.

pymtl3/passes/rtlir/behavioral/BehavioralRTLIRTypeCheckL3Pass.py
```diff
@@ -82,32 +81,29 @@ def visit_StructInst( s, node ):
       type_instance = cls()
     except TypeError:
       raise PyMTLTypeError( s.blk, node.ast,
-""""\
-__init__ of BitStruct {} should take 0 arguments! You can achieve this by
+f""""\
```
yo96: four `"` here?

pymtl3/passes/sverilog/import_/ImportPass.py
```diff
@@ -567,7 +566,7 @@ def gen_signal_init_c( s, name, port ):
           ret[ idx ] = "  " + ret[ idx ]
 
     else:
-      ret.append('m->{name} = {deference}model->{name};'.format(**locals()))
+      ret.append(f'm->{name} = {deference}model->{name};')
```
yo96: much cleaner!

jsn1993: yeah most of these **locals are gone ... since they basically serve the same purpose as fstring

PR #100, opened 2019-10-05
elaboration performance improvement

jsn1993:
Ufortunately for HUGE designs I cannot really do much because there are 30k+ update blocks and the elaboration code involves a TON of for/if/isinstance statements, rending itself JIT-unfriendly


jsn1993:
how do we actually fix those travis error due to hypothesis/attr?

pymtl3/dsl/Component.py
```diff
@@ -73,7 +73,7 @@ def _collect_objects_local( s, filt ):
     stack = []
     for (name, obj) in s.__dict__.items():
       if   isinstance( name, str ):
-        if not name.startswith("_"): # filter private variables
+        if name[0] != '_': # filter private variables
```
ptpan: can name be an empty string?

jsn1993: i don't think so

pymtl3/dsl/Component.py
```diff
@@ -490,15 +506,17 @@ def unlock_simulation( s ):
     except:
       raise AttributeError("Cannot unlock an unlocked/never locked model.")
 
+    swapped_values  = defaultdict(list)
```
ptpan: Is this mostly a performance optimization that tries to remove unnecessary attribute lookups?

requirements.txt
```diff
@@ -1,9 +1,7 @@
 pytest >= 4.4
 hypothesis >= 4.18.3
-pytest-xdist
 cffi
 greenlet
-pyparsing
 graphviz
```
ptpan: Cool. Let's definitely do more clean up like this to make pip install much faster

cbatten:
definitely! making pip install faster is a great goal

pymtl3/dsl/Component.py
jsn1993: Yes ... I try to only do it when it doesn't hurt readability ...

requirements.txt
jsn1993: I wonder if we can remove graphviz

ptpan: IMO graphviz really should be a developer tool that facilitates debugging. We probably should have different "dependency groups" which expose different set of tools to different audience. For example the "product" group should have core dependencies only, whereas "dev" group might have extra debugging facilitates like graphviz. Python formatters and code coverage tools should probably also go to the "dev" group.

jsn1993: hmm maybe we can move "CI dependencies" to .travis.yml?

PR #101, opened 2019-10-15
PrintWavePass + CollectSignalPass

kaishuocheng:
I have created printwavepass and collectsignalPass, along with their tests.

 In a word, collectsignalpass collects signals and stored it as an attribute _collectsignals of "top". PrintWavePass adds a method "printwave" to top. When called, top.printwave(top) will work. More details are explained in two corresponding tests.

Passgroup is slightly modified, importing two passes, and add them to "SimulationPass" block.


cbatten:
Can we call it `print_wave` instead of `printwave`?

kaishuocheng:
> Can we call it `print_wave` instead of `printwave`?

Yeap, I will change the name.

pymtl3/passes/CollectSignalPass.py
```diff
@@ -0,0 +1,131 @@
+"""
+========================================================================
+CollectSignalPass.py
+========================================================================
+#collects signals and stored it as _collectsignals attribute of top.
```
ptpan: The lines included inside `"""` is treated as if they are multi-line comments so the `#` at the beginning is not necessary.

pymtl3/passes/CollectSignalPass.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if hasattr( top, "_cl_trace" ):
+      schedule = top._cl_trace.schedule
+    else:
+      schedule = top._sched.schedule
+
+    top._wav = PassMetadata()
+
+    schedule.append( self.collect_sig_func( top, top._wav ) )
+  def collect_sig_func( self, top, wavmeta ):
```
ptpan: Should be a blank line before this since they belong to different methods

pymtl3/passes/CollectSignalPass.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # Give all ' and " characters a preceding backslash for .format
+    wav_srcs = []
+    for i, net in enumerate( trimmed_value_nets ):
+      if i != wavmeta.clock_net_idx:
+        wav_srcs.append( dump_wav_per_signal.format( i, net[0]) )
+    deepcopy # I have to do this to circumvent the tools
+    wavmeta.sigs = {}
+    char_length = 5
+#signals are stored in top.signals
```
ptpan: Comments should be aligned with their blocks too

pymtl3/passes/PassGroups.py
```diff
@@ -43,8 +45,11 @@
   DynamicSchedulePass(),
   CLLineTracePass(),
   VcdGenerationPass(),
+  #WaveGenPass(),
```
ptpan: If `WaveGenPass` is not critical to the functionality of this PR, we probably should remove this?

pymtl3/passes/PassGroups.py
```diff
@@ -43,8 +45,11 @@
...
+  CollectSignalPass(),
+  PrintWavePass(),
   SimpleTickPass(),
-  Component.lock_in_simulation
+Component.lock_in_simulation
```
ptpan: Can you take a pass to make sure lines like this are properly aligned? We would like to make sure the codebase has a consistent style by checking each PR in details...

pymtl3/passes/test/VcdGenPass_test.py
```diff
@@ -0,0 +1,62 @@
+#=========================================================================
+# VcdGenPass_test.py
```
ptpan: I guess we probably should remove this file because it was meant to be a toy/starting point to the development, and not a well-formed test case

pymtl3/passes/WaveGenPass.py
```diff
@@ -0,0 +1,189 @@
...
+import py
+import sys
+from pymtl3.dsl import Const
+from pymtl3.passes.BasePass import BasePass, PassMetadata
+
+from .errors import PassOrderError
+
+
+
+class WaveGenPass( BasePass ):
```
ptpan: What is the difference between `WaveGenPass` and `SignalCollectPass`? Only the latter pass is included in the pass group so `WaveGenPass` seems to be unused?

pymtl3/passes/PrintWavePass.py
```diff
@@ -0,0 +1,88 @@
...
+from pymtl3.dsl import Const
+from pymtl3.passes.BasePass import BasePass, PassMetadata
+
+
+class PrintWavePass( BasePass ):
+
+  def __call__( self, top ):
+    setattr(top,"printwave",helpprint)
+
+def helpprint(self):
```
ptpan: The current codebase uses `underscore_separated_lower_case_names` for functions, methods, and variable names. I think new PR's should also adopt this style for consistency.

ptpan:
@jsn1993 Do you think we should have a different simulation pass group for CLI wave generation? I guess we can manually enable this in demos but keeping normal execution the same as before.

cbatten:
Not sure ... something to think about. I think it should be analogous to VCD Gen. Is that pass on by default? Wavegen should not be on by default because it is not meant for long running simulations since it saves all of the signal trace data for post processing ...

cbatten:
I have an idea on how to apply passes from the command line. I really want to be able to do something like this:

ptpan:
> I think it should be analogous to VCD Gen. Is that pass on by default?

No we have a different pass group for that.

cbatten:
Right but the question is how do we turn on that pass group? I think we want to avoid having to make explicit py.test command line options for every variant of pass groups we want to apply for a simulation ... thus my idea of a single `--apply` command line option for py.test?

pymtl3/passes/WaveGenPass.py
kaishuocheng: Yeah WaveGenPass should not be included.  I guess I should cancel some push changes?

ptpan:
> Right but the question is how do we turn on that pass group? I think we want to avoid having to make explicit py.test command line options for every variant of pass groups we want to apply for a simulation ... thus my idea of a single `--apply` command line option for py.test?

This kind of reminds me the `opt` tool from LLVM which allows you to apply one pass on a input program. I guess we still want a "common" pass group which builds up the component that other passes work on, and then applying one interesting pass from command line input.

pymtl3/passes/WaveGenPass.py
ptpan: It's OK if you decide some of the code should not be included -- just remove them in a new commit!

pymtl3/passes/test/VcdGenPass_test.py
kaishuocheng: Can you also take a look at CollectSignalPass_test? It follows the style of this VcdGenPass, and I not sure if it is sufficient.

pymtl3/passes/CollectSignalPass.py
kaishuocheng: Also as a general question, how do I make changes to this pull request? If I change the branch, would changes also carried in this pull request?

cbatten:
yes! we would have a common base pass group -- I actually think the `--apply` pass groups would be applied _before_ the common simulation pass group ...

pymtl3/passes/CollectSignalPass.py
ptpan: You can keep pushing commits to this branch, and the changes will be updated here

pymtl3/passes/test/VcdGenPass_test.py
ptpan: I think passes like VcdGen and WaveGen are hard to test partly due to the lack of accurate reference output. The "toy" case itself is fine because it works as a simple directed test if we are able to somehow check the results of `SignalCollectPass` or `PrintWavePass`. Maybe asserting the names of signals appear in the output?

kaishuocheng: Yeap, that works

pymtl3/passes/CollectSignalPass.py
kaishuocheng: Got it!

pymtl3/passes/PrintWavePass.py
```diff
@@ -0,0 +1,127 @@
...
+            if sig != "s.clk" and sig != "s.reset":
+                maxlength = thislength
+
+    print(" "*(maxlength+1),end = "")
+
+    for i in range(0,len(allsignals["s.clk"]),5):
+          print(_tick + str(i)+ " "*(5*char_length-1) ,end="")
+
+    #signals
+    for sig in allsignals:
```
jsn1993: we usually put underscore between words. For example I would suggest ```all_signals``` here

pymtl3/passes/PrintWavePass.py
```diff
@@ -0,0 +1,127 @@
...
+import py
+import sys
+from pymtl3.dsl import Const
+from pymtl3.passes.BasePass import BasePass, PassMetadata
+import random
+
+class PrintWavePass( BasePass ):
+
+  def __call__( self, top ):
+    setattr(top,"_print_wave",_help_print)
```
jsn1993: might not need to use setattr here because _print_wave is a constant string

pymtl3/passes/PrintWavePass.py
jsn1993: I guess you can take a pass through the whole PR and improve the variable names

kaishuocheng: Got it, I will work on that

pymtl3/passes/PrintWavePass.py
kaishuocheng: I am trying to add a method to the top object, so it can call top._print_wave(top) to print out. 
Is there a better way to do this?

jsn1993: You can just do ```top._print_wave = _help_print```.
setattr is usually used when the attribute string is not a constant string

kaishuocheng: Ah, I see. Thanks!

jsn1993:
Kaishuo, just to get you prepared, we will be pretty tough on this review since it's the first time you submit a PR to brg organization. We not only want to maintain high code quality in this repo, but also want to share with you what the best practices in our opinion are.

kaishuocheng:
> Kaishuo, just to get you prepared, we will be pretty tough on this review since it's the first time you submit a PR to brg organization. We not only want to maintain high code quality in this repo, but also want to share with you what the best practices in our opinion are.

Yeap, no worries. I totally understand. Personally I also think this is a good practice to help me learn about what "real-life" coding is. Most of my coding before are for assignments and tests lol. Yeah no worries, feedbacks of all types are welcome.

pymtl3/passes/CollectSignalPass.py
```diff
@@ -0,0 +1,132 @@
...
+========================================================================
+CollectSignalPass.py
+========================================================================
+collects signals and stored it as _collect_signals attribute of top.
+
+Author : Kaishuo Cheng
+Date   : Oct 4, 2019
+"""
+
+from __future__ import absolute_import, division, print_function
```
jsn1993: Remove this

pymtl3/passes/CollectSignalPass.py
```diff
@@ -0,0 +1,132 @@
...
+import time
```
jsn1993: apply tools

pymtl3/passes/test/PrintWavePass_test.py
```diff
@@ -0,0 +1,63 @@
...
+  dut = Toy()
+  dut.elaborate()
+
+  # Setup the simulation
+  dut.apply( SimulationPass )
+  dut.sim_reset()
+  # Test vector
+  vector = [
+    #  i        inlong       out
+    b32(1),    b32(2),    b32(3),
```
jsn1993: Align each column

pymtl3/passes/test/CollectSignalPass_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+def processbinary(sig):
+    """
+    Returns int value from a signal in 32b form. Used for testing.
+    Example: input: 0b00000000000000000000000000000000
+             output: 0
+    """
+    tempint = int(sig[2:],2)
+    if sig[2] == '1':
+        return tempint -2 **32      #taking 2's complement.
```
jsn1993: Put the two lines of comments before return

pymtl3/passes/test/CollectSignalPass_test.py
```diff
@@ -0,0 +1,90 @@
...
+    s.in1 = InPort( Bits32 )
+    s.out = OutPort( Bits32 )
+
+    @s.update
+    def add_upblk():
+      # This update block models the behavior of a 32-bit adder
+      s.out = s.in0 + s.in1
+
+def processbinary(sig):
+    """
```
jsn1993: Take a pass to use 2-space indentation throughout all the files

jsn1993:
yeow @ptpan the tool was applied to examples too ...

cbatten:
What is the current state of this pull request? Can we try and get it merged this week? I would love to be able to use the text waveform generator in repl examples ...

jsn1993:
I will need merge master into this branch and clean it up.

cbatten:
Ok! Can Kai help?

jsn1993:
> Ok! Can Kai help?

I'm taking a pass to reorganize and optimize the code.

PR #102, opened 2019-10-23
Run examples on CI; improve requirement management

ptpan:
1. CI now runs the tests under `examples`
2. Organize the dependencies into three groups
- `release` is a core group of dependencies required for running PyMTL
- `dev` includes `release` and other tools that facilitate PyMTL development (`graphviz`, formatters, etc.)
- `CI` includes `dev` and tools required for performing CI (mainly code coverage plugins)
- PyMTL package on PyPI will also have updated dependencies: `pytest-xdist`, `pyparsing`, and `graphviz` are removed


jsn1993:
Does this change whatever we need to do currently? Can we still ```pip install pymtl3```? Does it affect travis? How do we specify when we install it as dev?

ptpan:
> Does this change whatever we need to do currently? Can we still `pip install pymtl3`? Does it affect travis? How do we specify when we install it as dev?

1. We can still pip install pymtl3. After this PR and another PyPI release, `pymtl3` will not longer depend on `graphviz`, `pytest-xdist`, `pyparsing`
2. Travis now uses dependencies specified in `CI.txt` which is the same as before
3. After we git clone the repo, do `pip install -r requirements.txt && pip install --editable .`

jsn1993:
Great. I think the last thing we need to decide for this PR is whether we want to drop 3.6 or 3.7 in travis.

ptpan:
> Great. I think the last thing we need to decide for this PR is whether we want to drop 3.6 or 3.7 in travis.

Seems to me that having both 3.6 and 3.7 in CI is really helpful because we are actively using 3.7, and officially support 3.6. If the only concern for CI is speed, I'd say we should probably look at other ways...

jsn1993:
> > Great. I think the last thing we need to decide for this PR is whether we want to drop 3.6 or 3.7 in travis.
> 
> Seems to me that having both 3.6 and 3.7 in CI is really helpful because we are actively using 3.7, and officially support 3.6. If the only concern for CI is speed, I'd say we should probably look at other ways...

ok. I put pypy as the first launch, but it seems like the pypy run is failing ,..weird

jsn1993:
Man, adding example tests to CI revealed a lot of bugs introduced by the bit struct PR ......

cbatten:
We should create better "real" tests in the pymtl kernel then? Can we hypothesis to help stress test bitstructs? It is ok to catch bugs with examples, but that should really only be a last resort ... we should try and get pymtl kernel tests that would have failed too right?

PR #110, opened 2019-11-17
Make constant-sized slicings translatable

ptpan:
Constant-sized slicings (e.g. `s.out[v*8 : v*8 + 8]` where `v` could be a loop index) can be nicely translated to Verilog (using `+:`). This PR supports translating slices whose upper bound is larger (i.e. adding a constant) than its lower bound by a constant.

### DIFF
--- a/pymtl3/passes/rtlir/behavioral/test/BehavioralRTLIRPass_test.py
+++ b/pymtl3/passes/rtlir/behavioral/test/BehavioralRTLIRPass_test.py
@@ -425,6 +425,15 @@ def test_ff_upblk( do_test ):
         Assign( Attribute( Base( a ), 'out0' ), BinOp( Attribute( Base( a ), 'in0' ), Add(), Attribute( Base( a ), 'in1' ) ), False ),
         ] )
   }
+  a._test_vector = [
+                'in0                in1              *out',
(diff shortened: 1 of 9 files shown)


jsn1993:
wow nice. is this easy?

ptpan:
> wow nice. is this easy?

Not too hard. We had a buggy implementation before which did not include translation support. I fixed some bugs in it and added translation for it.

jsn1993:
> > wow nice. is this easy?
> 
> Not too hard. We had a buggy implementation before which did not include translation support. I fixed some bugs in it and added translation for it.

Great. Looks like it's not a lot of changes.

pymtl3/passes/rtlir/behavioral/test/BehavioralRTLIRPass_test.py
```diff
@@ -426,3 +426,29 @@ def update_ff_upblk():
         ] )
   }
   do_test( a )
+
+def test_fixed_size_slice( do_test ):
```
jsn1993: Is this test included in TranslationImport test?

ptpan: Not yet. I'll add it to translation tests

PR #111, opened 2019-11-30
Connect const struct to a signal

jsn1993:
This PR implements connecting a signal to a constant bitstruct. Basically I use repr(x) as the writer when generating connection block. Note that repr(x) cannot disambiguate A.X and B.X because they all look like X(Bits4(2)). I throw an error for this case. Otherwise it's good to go.

I also added tests for pymtl3.datatypes.helpers

@ptpan can you also add some translation tests based on the added tests in ComponentLevel_test.py

### DIFF
--- a/pymtl3/datatypes/helpers.py
+++ b/pymtl3/datatypes/helpers.py
@@ -85,15 +85,15 @@ def to_bits( obj ):
   assert is_bitstruct_inst( obj ), f"{obj} is not a valid PyMTL Bitstruct!"
   return concat( *[ to_bits(getattr(obj, v)) for v in obj.__bitstruct_fields__.keys() ] )
 
-def _get_bitstruct_inst_all_classes( obj ):
+def get_bitstruct_inst_all_classes( obj ):
   # list: put all types together
   if isinstance( obj, list ):
-    return functools.reduce( operator.or_, [ _get_bitstruct_inst_all_classes(x) for x in obj ] )
+    return functools.reduce( operator.or_, [ get_bitstruct_inst_all_classes(x) for x in obj ] )
   ret = { obj.__class__ }
   # BitsN or int
   if isinstance( obj, (Bits, int) ):
     return ret
   # BitStruct
   assert is_bitstruct_inst( obj ), f"{obj} is not a valid PyMTL Bitstruct!"
-  return ret | functools.reduce( operator.or_, [ _get_bitstruct_inst_all_classes(getattr(obj, v))
+  return ret | functools.reduce( operator.or_, [ get_bitstruct_inst_all_classes(getattr(obj, v))
                                                 for v in obj.__bitstruct_fields__.keys() ] )
--- a/pymtl3/datatypes/helpers.py
+++ b/pymtl3/datatypes/helpers.py
@@ -8,7 +8,9 @@ should be implemented in RPython.
 Author : Shunning Jiang
 Date   : Nov 3, 2017
 """
+import functools
 import math
+import operator
 
 from .bits_import import *
 from .bitstructs import is_bitstruct_class, is_bitstruct_inst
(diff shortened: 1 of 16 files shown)


pymtl3/datatypes/helpers.py
```diff
@@ -70,9 +75,25 @@ def get_nbits( Type ):
...
+    return concat( *[ to_bits(x) for x in obj ] )
+
   # BitsN
   if isinstance( obj, Bits ):
     return obj
   # BitStruct
   assert is_bitstruct_inst( obj ), f"{obj} is not a valid PyMTL Bitstruct!"
   return concat( *[ to_bits(getattr(obj, v)) for v in obj.__bitstruct_fields__.keys() ] )
+
+def _get_bitstruct_inst_all_classes( obj ):
```
ptpan: we probably should allow designers use this (in tests) if they want? Maybe remove the underscore?

pymtl3/datatypes/test/helpers_test.py
```diff
@@ -0,0 +1,82 @@
...
+helpers_test.py
+==========================================================================
+Test cases for helper functions
+
+Author : Shunning Jiang
+  Date : Nov 30, 2019
+"""
+
+from pymtl3.datatypes import *
+from pymtl3.datatypes.helpers import _get_bitstruct_inst_all_classes
```
ptpan: This helper is used here. If it appears in a test then we might want to expose it instead of prefixing it with `_`

pymtl3/dsl/test/ComponentLevel3_test.py
```diff
@@ -766,6 +775,108 @@ def construct( s ):
...
+
+  @bitstruct
+  class SomeMsg2:
+    a: A.SomeMsg1
+    b: B.SomeMsg1
+
+  class Top( ComponentLevel3 ):
+    def construct( s ):
+      s.out = OutPort(SomeMsg2)
+      connect( s.out, SomeMsg2(A.SomeMsg1(1,2),B.SomeMsg1(3,4)) )
```
ptpan: By the way, I'm thinking if we can organize most of our test cases by the test component used in it. To add translation test for the same feature I have to duplicate the test component again in another test file...

pymtl3/passes/GenDAGPass.py
```diff
@@ -46,19 +47,15 @@ def _generate_net_blocks( self, top ):
...
-    # Note that it introduced problems with multiple update blocks having
-    # the same const writer at the same hostobj. As a result, I add a
-    # suffix to each block like Bits1_xxx_no_12, Bits1_xxx_no_13 to
-    # disambiguate the blocks
-
-    hostobj_allsrc = defaultdict(str)
-    hostobj_bits   = defaultdict(set)
-    blkname_meta   = {}
-    blkname_suffix = {}
+    # TODO see if directly compiling AST instead of source can be faster
```
ptpan: Maybe have a comment here about why we need block-by-block compilation (to differentiate different structs with the same __name__)?

pymtl3/passes/yosys/translation/structural/YosysStructuralTranslatorL1.py
```diff
@@ -253,7 +253,8 @@ def rtlir_tr_current_comp( s, comp_id, comp_rtype ):
     s.deq.append( {'attr':[], 'index':[], 's_attr':"", 's_index':""} )
     return ''
 
-  def rtlir_tr_literal_number( s, nbits, value ):
+  def rtlir_tr_literal_number( s, nbits, value, first_called = True ):
```
ptpan: I'll think more about the plan for the yosys backend. Seems like it's becoming even more hacky...

pymtl3/datatypes/helpers.py
jsn1993: yeah we could

pymtl3/dsl/test/ComponentLevel3_test.py
jsn1993: I guess import a test component from DSL might not be a good idea, but reusing tests across backends should be good. This is because not all tests in DSL are Component test but ComponentLevel1,2,3 etc.

ptpan: Isn't a valid ComponentL1 test also a valid Component test?

PR #113, opened 2019-12-02
Add default argument Bits1 to Signal (Wire,InPort,OutPort)

jsn1993:
As per @cbatten suggested, this can be a really nice syntactic sugar when people model net-list-level stuff.

This test shows an example https://github.com/cornell-brg/pymtl3/commit/b7d58e7b1ca4045e9e9306dca8b836523a94b9cc#diff-93f90fe9147a2c1bc54ad23e9e2732fdR710

### DIFF
--- a/pymtl3/dsl/Connectable.py
+++ b/pymtl3/dsl/Connectable.py
@@ -10,7 +10,7 @@ Date   : Apr 16, 2018
 import types
 from collections import deque
 
-from pymtl3.datatypes import Bits, mk_bits
+from pymtl3.datatypes import Bits, Bits1, mk_bits
 
 from .errors import InvalidConnectionError
 from .NamedObject import DSLMetadata, NamedObject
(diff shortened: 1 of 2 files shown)


cbatten:
@yo96 @ptpan what do you think? Do you think this is a good idea?

ptpan:
I'm OK with this -- we have been doing similar things to facilitate common cases.

cbatten:
I think it is primarily useful when doing GL modeling like this:

```

ptpan:
> I think it is primarily useful when doing GL modeling like this:
> 
> ```
> class FullAdder( Component ):
>   def construct( s ):
>     s.a    = InPort ()
>     s.b    = InPort ()
>     s.cin  = InPort ()
>     s.sum  = OutPort()
>     s.cout = OutPort()
> 
>     s.sum  //= lambda: s.a ^ s.b ^ s.cin
>     s.cout //= lambda: (s.a & s.b) | (s.a & s.cin) | (s.b & s.cin)
> 
> class NAND( Component ):
>   def construct( s ):
>     s.a = InPort ()
>     s.b = InPort ()
>     s.y = OutPort()
> 
>     s.y //= lambda: ~( s.a & s.b )
> ```

The only concern I have is that I'm not sure if this works for static type checking (treat missing type parameters as some default type). Other than that I think that will be a convenient feature.

cbatten:
OK. Maybe we can add it and just keep in mind that you might not be able to use this syntatic sugar if you are using type checking ... which begs the question of whether or not you are going to push for getting MyPyMTL into master :)

PR #114, opened 2019-12-04
Reorganize pass folder structure

jsn1993:


jsn1993:
@ptpan can you push the verilator wrapper bug fix directly to this branch?

examples/ex02_cksum/test/ChecksumRTL_test.py
```diff
@@ -115,19 +116,16 @@ def setup_method( s, method ):
...
-      s.vcd_file_name = ""
+    s.vcd_file_name = ""
 
   def run_sim( s, th, max_cycles=1000 ):
 
     # Check for vcd dumping
     if s.vcd_file_name:
-      th.dump_vcd = True
-      th.vcd_file_name = s.vcd_file_name
+      th.config_tracing = TracingConfigs(tracing='text_fancy', vcd_file_name=s.vcd_file_name)
```
ptpan: If `TracingConfigs` are options for tracing in general, then this option probably should have names like `output_name` or something?

ptpan: `vcd` makes it too specific to a single option

pymtl3/passes/PassGroups.py
```diff
@@ -1,88 +1,56 @@
...
+from .sim.WrapGreenletPass import WrapGreenletPass
+from .tracing.CLLineTracePass import CLLineTracePass
+from .tracing.CollectSignalPass import CollectSignalPass
+from .tracing.LineTraceParamPass import LineTraceParamPass
+from .tracing.PrintWavePass import PrintWavePass
+from .tracing.VcdGenerationPass import VcdGenerationPass
+
+
+# SimpleSim can be used when the UDG is a DAG
+class SimpleSimPass( BasePass ):
```
ptpan: Now that each "PassGroup" is actually a pass, I wonder whether we should keep the name "PassGroup" or not... Is "Passes" better than "PassGroups"? What do you think?

pymtl3/passes/mamba/PassGroups.py
```diff
@@ -0,0 +1,27 @@
+from .HeuristicTopoPass import HeuristicTopoPass
+from .TraceBreakingSchedTickPass import TraceBreakingSchedTickPass
+from .UnrollTickPass import UnrollTickPass
+
+UnrollSim = [
```
ptpan: When we get this working, we need to make this consistent with the main pass group file.

pymtl3/passes/tracing/test/CollectSignalPass_test.py
```diff
@@ -25,31 +28,31 @@ def add_upblk():
       s.out = s.in0 + s.in1
 
 def processbinary(sig):
```
ptpan: Add an underscore between these two words?

examples/ex02_cksum/test/ChecksumRTL_test.py
jsn1993: only vcd pass will look at vcd_file_name, and other tracing passes probably have their own file name option because there can be cases where you want to dump textwave to one file and vcd to another file.

pymtl3/passes/PassGroups.py
jsn1993: Hmmm I mean it's basically a "top-level pass" or "a sequence of passes that serve a purpose"

pymtl3/passes/mamba/PassGroups.py
jsn1993: I can do it in this PR

pymtl3/passes/PassGroups.py
jsn1993: @cbatten comments on this are welcome

pymtl3/passes/tracing/test/CollectSignalPass_test.py
jsn1993: done

PR #115, opened 2019-12-08
eval comb and schedule pass refactoring

jsn1993:
In this PR I did the following:

1. I added eval_combinational to the component in SimpleTickPass when there is no method port in the design (I think this is an OK indicator of whether there are CL modules inside the DUT for now. Later we can also unmark update_once).

2. To accomplish 1., I need to actually split intra-cycle schedule from update_ff and posedge. I went ahead and abandoned the previous tick assembling scheme where each pass adds to a list. Then I reimplemented all the passes that adds a function to the schedule like vcd signal collections and such to create a function in their own scheme. In SimpleTickPass or other tick generation passes, I assemble all the functions from these passes with the schedule and ff with the desired order.

3. Then I actually need to tweak all the places where the simulation is done in a similar flavor to TestVectorSimulator or directly uses TestVectorSimulator. Now we have the same TestVectorSimulator implementation as v2.

### DIFF
--- a/examples/ex02_cksum/test/ChecksumRTL_test.py
+++ b/examples/ex02_cksum/test/ChecksumRTL_test.py
@@ -54,16 +54,19 @@ def checksum_rtl( words ):
   dut.send.rdy = b1(1)
   while not dut.recv.rdy:
     dut.recv.en = b1(0)
+    dut.eval_combinational()
     dut.tick()
 
   # Feed in the input
   dut.recv.en = b1(1)
   dut.recv.msg = bits_in
+  dut.eval_combinational()
   dut.tick()
 
   # Wait until the checksum unit is about to send the message
   while not dut.send.en:
     dut.recv.en = b1(0)
+    dut.eval_combinational()
     dut.tick()
 
   # Return the result
--- a/pymtl3/passes/sim/SimpleTickPass.py
+++ b/pymtl3/passes/sim/SimpleTickPass.py
@@ -12,6 +12,12 @@ from pymtl3.dsl.errors import UpblkCyclicError
 from pymtl3.passes.BasePass import BasePass
 from pymtl3.passes.errors import PassOrderError
 
+# Shunning: We are aware of the problem that there may be multiple places
+# that assembles the function for the user to tick the simulator.
+# I think LLVM applies passes based on the order they are inserted to the
+# pass manager. LLVM also has some local ordering mechanism to figure out
+# the dependencies between passes. Currently we haven't got to the point
+# where we have enough passes to manage, but we should keep this in mind.
 
 class SimpleTickPass( BasePass ):
 
--- a/pymtl3/passes/sim/SimpleTickPass.py
+++ b/pymtl3/passes/sim/SimpleTickPass.py
@@ -7,6 +7,7 @@ Generate a simple tick function (no Mamba techniques here)
 Author : Shunning Jiang
 Date   : Dec 26, 2018
 """
+from pymtl3.dsl import MethodPort
 from pymtl3.dsl.errors import UpblkCyclicError
 from pymtl3.passes.BasePass import BasePass
 from pymtl3.passes.errors import PassOrderError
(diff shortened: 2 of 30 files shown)


examples/ex02_cksum/test/ChecksumRTL_test.py
```diff
@@ -54,16 +54,19 @@ def checksum_rtl( words ):
   dut.send.rdy = b1(1)
   while not dut.recv.rdy:
     dut.recv.en = b1(0)
+    dut.eval_combinational()
```
ptpan: Just curious: in what kind of situations would designers want to call `eval_combinational` separately from a `tick`?

jsn1993: when you modify some signals and you want to let these changes immediately make effect before calling tick

pymtl3/passes/sim/SimpleTickPass.py
```diff
@@ -22,13 +23,44 @@ def iterative():
...
+    # Currently the tick order is:
+    # [ ff, tracing, posedge, clear_cl_trace, update ]
+
+    final_schedule = []
+    # call ff blocks first
+    final_schedule.extend( top._sched.schedule_ff )
+
+    # append tracing related work
 
+    if hasattr( top, "_tracing" ):
```
ptpan: What happens if both OpenLoopCLPass and VcdTracingPass are applied? Will the functions be inserted to _sched twice?

ptpan:
My main concern about this PR is whether we want a systematic way to assembly the final schedule. Currently two passes (`OpenLoopCL` and `SimpleTickPass`) have to hardcode global ordering of functions that appear in the final schedule. I guess if we see more passes like these two coming then probably some form of local ordering mechanism will be really helpful. Thoughts? @jsn1993

jsn1993:
> My main concern about this PR is whether we want a systematic way to assembly the final schedule. Currently two passes (`OpenLoopCL` and `SimpleTickPass`) have to hardcode global ordering of functions that appear in the final schedule. I guess if we see more passes like these two coming then probably some form of local ordering mechanism will be really helpful. Thoughts? @jsn1993

To be honest, I have the same concern. I think LLVM apply passes based on the order they are inserted to the pass manager, and they also have some local ordering mechanism to figure out the dependencies between passes. I guess we haven't got to the point where we have enough passes to manage. For OpenLoopCL and SimpleTick, they are two different simulation mechanisms so they might want ot have different global ordering so I think that should be fine.

pymtl3/passes/sim/SimpleTickPass.py
jsn1993: No, the second VcdGenerationPass call will overwrite the previous function

ptpan:
> > My main concern about this PR is whether we want a systematic way to assembly the final schedule. Currently two passes (`OpenLoopCL` and `SimpleTickPass`) have to hardcode global ordering of functions that appear in the final schedule. I guess if we see more passes like these two coming then probably some form of local ordering mechanism will be really helpful. Thoughts? @jsn1993
> 
> To be honest, I have the same concern. I think LLVM apply passes based on the order they are inserted to the pass manager, and they also have some local ordering mechanism to figure out the dependencies between passes. I guess we haven't got to the point where we have enough passes to manage. For OpenLoopCL and SimpleTick, they are two different simulation mechanisms so they might want ot have different global ordering so I think that should be fine.

Okay. Can you add a comment to these two passes saying we might want to improve this so that we don't complete forget about them?

jsn1993:
> > > My main concern about this PR is whether we want a systematic way to assembly the final schedule. Currently two passes (`OpenLoopCL` and `SimpleTickPass`) have to hardcode global ordering of functions that appear in the final schedule. I guess if we see more passes like these two coming then probably some form of local ordering mechanism will be really helpful. Thoughts? @jsn1993
> > 
> > 
> > To be honest, I have the same concern. I think LLVM apply passes based on the order they are inserted to the pass manager, and they also have some local ordering mechanism to figure out the dependencies between passes. I guess we haven't got to the point where we have enough passes to manage. For OpenLoopCL and SimpleTick, they are two different simulation mechanisms so they might want ot have different global ordering so I think that should be fine.
> 
> Okay. Can you add a comment to these two passes saying we might want to improve this so that we don't complete forget about them?

done

PR #116, opened 2019-12-09
Revamp Bits strategy and implement Bitstruct strategy

jsn1993:
Now I think we should create strategies in a generator-like style. At strategy construction time, our bitwidth/bitstruct types are already fixed, which means all the arguments won't change during runtime. As a result, we should be able to capture them in closure and return a strategy with minimum metadata access like object dict access or creating types.

Based on these, I reimplemented Bits strategy and added custom min_value/max_value support since there can be cases where the full range of value doesn't make sense. I also write a bitstruct strategy based on recursive construction of strategy list, and the real strategy is very succinct and simple.

For complete functionality, I think I still need to implement setting min/max value limit from the top-level bitstruct using a trie-like data structure. I open the PR early to get some feedback on style.

### DIFF
--- a/pymtl3/datatypes/strategies.py
+++ b/pymtl3/datatypes/strategies.py
@@ -4,30 +4,96 @@ strategies.py
 ==========================================================================
 Hypothesis strategies for built-in data types.
 
-Author : Yanghui Ou
-  Date : June 16, 2019
+Author : Shunning Jiang, Yanghui Ou
+  Date : Dec 9, 2019
 """
 import hypothesis
 from hypothesis import strategies as st
 
-from .bits_import import mk_bits
+from .bits_import import Bits, mk_bits
+from .bitstructs import is_bitstruct_class
 
 #-------------------------------------------------------------------------
 # A generic bits strategy.
 #-------------------------------------------------------------------------
 
-@st.composite
-def bits( draw, nbits, signed=False ):
-  if nbits == 1:
-    value = draw( st.booleans() )
-  elif not signed:
-    min_value = 0
-    max_value = 2**nbits - 1
-    value = draw( st.integers(min_value, max_value) )
-  else:
-    min_value = - ( 2 ** (nbits-1) )
-    max_value = 2**(nbits-1) - 1
-    value = draw( st.integers(min_value, max_value) )
+# Return the SearchStrategy for Bits_nbits with the support to limit
+# min/max value and setting signedness
 
+def bits( nbits, signed=False, min_value=None, max_value=None ):
   BitsN = mk_bits( nbits )
-  return BitsN( value )
+
+  if (min_value is not None or max_value is not None) and signed:
+    raise ValueError("bits strategy currently doesn't support setting "
+                     "signedness and min/max value at the same time")
+
+  if min_value is None:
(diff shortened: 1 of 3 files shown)


jsn1993:
Setting manual min/max limit already implemented. Feature complete now! @yo96 take a look when you have time. @icyxiao96 you can take a look if you want

cbatten:
This is a great idea!

pymtl3/datatypes/strategies.py
```diff
@@ -4,30 +4,114 @@
...
+
+  return strategy_list() # RETURN A STRATEGY INSTEAD OF FUNCTION
+
+#-------------------------------------------------------------------------
+# strategies.bitstruct
+#-------------------------------------------------------------------------
+# Return the SearchStrategy for bitstruct type T with the support of
+# dictionary-based min/max value limit
+
+def bitstruct( T, limit_dict=None ):
```
yo96: bitstructs

pymtl3/datatypes/strategies.py
```diff
@@ -4,30 +4,114 @@
...
+
+  return strategy_bits() # RETURN A STRATEGY INSTEAD OF FUNCTION
+
+#-------------------------------------------------------------------------
+# strategies.bitslist
+#-------------------------------------------------------------------------
+# Return the SearchStrategy for a list of Bits with the support of
+# dictionary based min/max value limit
+
+def bitslist( types, limit_dict=None ):
```
yo96: bitslists

cbatten: I think it should be bitslist ... it is a list of bits not multiple lists of bits?

pymtl3/datatypes/test/strategies_test.py
```diff
@@ -37,4 +40,172 @@ def test_signed( nbits ):
...
+    assert 0xf0 <= bs.p2.y <= 0xff
+    print( bs )
+
+  print("-"*10,NestedPoint,"-"*10)
+  actual_test()
+
+def test_nnested_point_limited():
+  limit_dict = {
+    'p1': {
+      'x': (0xe0,0xef),
```
yo96: we should support user defined strategies here too

pymtl3/datatypes/strategies.py
yo96: well in hypothesis it's also st.lists not list, because it's a strategy that generate different lists.

cbatten: Ah ha -- I stand corrected :) we should definitely follow the hypothesis naming conventions.

pymtl3/datatypes/test/strategies_test.py
jsn1993: Done. Check the latest tests!

pymtl3/datatypes/test/strategies_test.py
```diff
@@ -25,10 +26,12 @@ def test_unsigned( nbits ):
   def actual_test( bits ):
     assert bits.nbits == nbits
     print( bits, bits.uint() )
+
+  print("-"*10,nbits,"-"*10)
```
yo96: maybe we can get rid of these print?

PR #117, opened 2019-12-12
Generic Method Interface and hypothesis stateful

jsn1993:
I started off implementing Callee/CallerIfcRTL/CL/FL in dsl as first-class citizens. Caller/Callee RTL, as the base class, is fully parametrized to support various method interfaces. Putting RTL/CL/FL Caller/Callee interfaces in dsl will make a lot of things easier. Also it was very inelegant previously, as we have CL and FL interfaces in dsl as Nonblocking and Blocking interfaces but all the RTL ones are created in an ad-hoc way in stdlib.

```
class CallerIfcRTL( CallIfcRTL ):
  def construct( s, en=None, rdy=None, MsgType=None, RetType=None ):
    < add en/rdy/msg/ret based on parameters>
```

 All these send/recv/get/give will inherit from these base Callee/Caller interfaces and construct with parameters.

```
class GetIfcRTL( CallerIfcRTL ):
  def construct( s, Type ):
    super().construct( en=True, rdy=True, MsgType=None, RetType=Type )
```

Then I went ahead and merge Yixiao's hypothesis stateful PR #54 into the branch and performed a massive refactoring process. Yanghui was doing some refactoring in that branch but he didn't finish. Since I already merge the original PR into this PR, we will close the original PR after we merge this PR in.


jsn1993:
Finally the generic method interface PR is merged in. Now this PR looks much better and we can start to brainstorm where to put those stateful tests!

ptpan:
I'm not sure if we still want this as part of the 4.0 release... @yo96 thoughts on this? If this has been inactive for way too long maybe we should close this?

ptpan:
Closing this now due to inactivity

PR #121, opened 2019-12-20
Placement pass initial pull request

tancheng:
Placement pass is a pass that hierarchically calculates the dimensional information (x, y, width, height) of each module for the final chip layout.
- There is a place() function taking width and height as parameters (or no parameters) in each module (or in the specific module you want to place).
- Users should split the area of the top module into grid/cells, then place/fill each submodule into the cell.
- The size of the placed top module is calculated hierarchically from the bottom (can even be the size of the standard cell) to top. The cells in the same hierarchy level have the same size, while the modules held by the cells could be in different sizes. 

Next step:
- Need to port to the latest pymtl3.
- Should include the area information about standard cells.

### DIFF
--- a/pymtl3/dsl/Component.py
+++ b/pymtl3/dsl/Component.py
@@ -30,6 +30,12 @@ class Component( ComponentLevel7 ):
       s.clk   = InPort( Bits1 )
       s.reset = InPort( Bits1 )
 
+      # physical dimension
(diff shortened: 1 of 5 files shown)


ptpan:
Closing this because the placement pass still seems too experimental and has been inactive for too long. Feel free to reopen this if we decide to explore this idea more.

PR #122, opened 2019-12-23
Refactor test data into centralized test directories

ptpan:
Previously the test data (components, test vectors, etc.) were included inside each test cases. This makes sharing test cases across multiple places (e.g., sharing test components and test vectors between different translation backends) challenging. 

This PR implement adds centralized test directories which include such test data to facilitate test reuse. Specifically, the top level `testcases` directory includes test components, test vectors, test_in functions and test_out functions. Each translation backend (`generic`, `sverilog`, and `yosys`) holds a different local `testcases` directory which copies the test cases from the top level directory and adds backend-specific data to them.


jsn1993:
Do you think we should put testcases folder in pymtl3/passes/ or the current pymtl3/?

ptpan:
> Do you think we should put testcases folder in pymtl3/passes/ or the current pymtl3/?

Oh I'll move it to pymtl3/passes/ for now.

ptpan:
> Wait, I was just asking. Is there any other places that we can use these test cases?

No. I initially put them there because mypy has a top level `testdata` folder. I guess this does not fit our case...

pymtl3/passes/backends/generic/test/RTLIRTranslator_L1_cases_test.py
```diff
@@ -5,59 +5,52 @@
...
-from pymtl3.datatypes import Bits1, Bits32
-from pymtl3.dsl import Component, InPort, OutPort, connect
-from pymtl3.passes.rtlir.util.test_utility import do_test, expected_failure
+import pytest
 
-from ..behavioral.test.BehavioralTranslatorL1_test import *
+from pymtl3.passes.rtlir.util.test_utility import expected_failure, get_parameter
+
+from ..behavioral.test.BehavioralTranslatorL1_test import (
+    test_generic_behavioral_L1 as behavioral,
```
jsn1993: Do you really want this "as" here? As far as I've seen they are only used once, so I think just import that test and use it below seems fine.

jsn1993:
Also I haven't seen this usage of AliasOf. Where is the no-gnu-unique test?

pymtl3/passes/backends/generic/test/RTLIRTranslator_L1_cases_test.py
ptpan: Right they are used only once as a short hand..

ptpan:
> Also I haven't seen this usage of AliasOf. Where is the no-gnu-unique test?

That test is at https://github.com/cornell-brg/pymtl3/blob/dba9fbcef70c93ced579310c22314a49b93e2db7/pymtl3/passes/backends/sverilog/test/TranslationImport_dynlib_close_test.py#L33 but I don't think it requires `AliasOf`... The one that potentially uses aliases is the hash of component name one. Currently it is implemented by adding multiple parameters to `DUT` which makes the name longer than 64 characters.

I do think `AliasOf` may be potentially useful to some test writers. Also it can be used to provide a mechanism that maps `TEST_VECTOR`, `TV_IN`, `TV_OUT`, etc. to a different name, which also can be useful.

pymtl3/passes/backends/generic/test/RTLIRTranslator_L1_cases_test.py
```diff
@@ -5,59 +5,49 @@
 # Date   : May 23, 2019
 """Test the RTLIR transaltor."""
```
jsn1993: typo

pymtl3/passes/backends/generic/test/RTLIRTranslator_L4_cases_test.py
```diff
@@ -5,18 +5,26 @@
 # Date   : May 23, 2019
 """Test the RTLIR transaltor."""
```
jsn1993: typo. Take a pass

PR #127, opened 2020-01-14
Fix a bug in the checker of `vl_src` import option

ptpan:
1. Fix an invalid name reference in the checker of `vl_src`
2. Improve the port map checker. Now you can only specify the names of ports that is different from the Verilog module you try to import (see `test_reg_incomplete_portmap`).

### DIFF
--- a/pymtl3/passes/backends/sverilog/import_/ImportConfigs.py
+++ b/pymtl3/passes/backends/sverilog/import_/ImportConfigs.py
@@ -213,6 +213,14 @@ class ImportConfigs( BasePassConfigs ):
   # Public APIs
   #-----------------------------------------------------------------------
 
+  # Override
+  def check( s ):
+    super().check()
+    
+    if not s.vl_flist and not os.path.isfile(expand(s.vl_src)):
+      raise InvalidPassOptionValue( 'vl_src', s.vl_src, s.PassName,
+                'vl_src should be a path to a file when vl_flist is empty!' )
+
   def create_vl_cmd( s ):
     top_module  =  f"--top-module {s.top_module}"
     src         = s.vl_src
@@ -358,7 +366,7 @@ class ImportConfigs( BasePassConfigs ):
                isinstance(p.get_dtype(), rdt.Vector) for n, p in all_ports), \
         f"Port map option currently requires all ports of {rtype.get_name()}"\
         f" to be a single vector port."
-    for name, _ in pm.items():
+    for name in pm.keys():
       if name not in all_port_names:
         raise InvalidPassOptionValue("port_map", pm, s.PassName,
           f"Port {name} does not exist in component {rtype.get_name()}!")
--- a/pymtl3/passes/backends/sverilog/import_/ImportConfigs.py
+++ b/pymtl3/passes/backends/sverilog/import_/ImportConfigs.py
@@ -162,8 +162,8 @@ class ImportConfigs( BasePassConfigs ):
 
     "port_map": Checker( lambda v: isinstance(v, dict), "expects a dict"),
 
-    "vl_src": Checker( lambda v: isinstance(v, str) and os.path.isfile(expand(v)) or \
-                s.vl_flist, "vl_src should be a path to a file when vl_flist is empty" ),
+    "vl_src": Checker( lambda v: isinstance(v, str) and (os.path.isfile(expand(v)) or not v),
+                "vl_src should be a path to a file or an empty string!" ),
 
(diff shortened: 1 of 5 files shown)


cbatten:
Only specifying the names of the ports which are different sounds like a nice improvement! Makes the simple case easy (names are the same) and the complex case possible (names are different). Does this change what this simple example looks like?

- https://gist.github.com/ptpan/fb8aae097af6e5a75679b5307e0b7426

ptpan:
> Only specifying the names of the ports which are different sounds like a nice improvement! Makes the simple case easy (names are the same) and the complex case possible (names are different). Does this change what this simple example looks like?
> 
> * https://gist.github.com/ptpan/fb8aae097af6e5a75679b5307e0b7426

No. In the gist example the placeholder has the same port names as the Verilog module and does not require a port map. One scenario in which this improvement is useful is when you import a Verilog module that has different names for clock and reset than `clk` and `reset` -- only `clk` and `reset` need to be mapped if you keep all other port names the same.

pymtl3/passes/backends/sverilog/import_/ImportConfigs.py
```diff
@@ -350,20 +350,18 @@ def vprint( s, msg, nspaces = 0, use_fill = False ):
...
     pm = s.port_map
     all_ports = rtype.get_ports_packed()
+    all_port_names = list(map(lambda x: x[0], rtype.get_ports_packed()))
     assert all(isinstance(p, rt.Port) and \
                isinstance(p.get_dtype(), rdt.Vector) for n, p in all_ports), \
         f"Port map option currently requires all ports of {rtype.get_name()}"\
         f" to be a single vector port."
-    for name, port in all_ports:
-      if name not in pm:
+    for name, _ in pm.items():
```
jsn1993: .keys()

pymtl3/passes/backends/sverilog/import_/test/VAdder.sv
```diff
@@ -1,12 +1,12 @@
 module VAdder #( parameter nbits = 32 )
 (
-  input clk,
-  input reset,
```
jsn1993: This worked?

pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
```diff
@@ -55,6 +55,36 @@ def construct( s ):
...
+  def tv_out( m, test_vector ):
+    if test_vector[1] != '*':
+      assert m.out == Bits32( test_vector[1] )
+  class VReg( Component ):
+    def construct( s ):
+      s.in_ = InPort( Bits32 )
+      s.out = OutPort( Bits32 )
+      s.config_sverilog_import = ImportConfigs(
+          vl_src = get_dir(__file__)+'VReg.sv',
+          port_map = {
```
jsn1993: I wonder if we should use the port object s.in_ instead of name in_

pymtl3/passes/backends/sverilog/import_/test/VAdder.sv
ptpan: The arguments passed to `construct` of a placeholder will be treated the parameters of the Verilog module. If the parameter is not specified, then the default value will be used.

pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
ptpan: We could. They have about the same number of characters so I'm wondering what would be the benefit of using port objects instead of names...

pymtl3/passes/backends/sverilog/import_/test/VAdder.sv
jsn1993: oh sorry i mean ```input clk``` instead of ```input logic clk```

pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
jsn1993: it could throw better errors earlier when the user specifies the key using an actual object i guess?

jsn1993: We can do this later this month

pymtl3/passes/backends/sverilog/import_/test/VAdder.sv
ptpan: This is a good point. This was supported by Verilator 4.008 and before but becomes an error in the most recent version of Verilator. We haven't updated the Verilator cached of pymtl3 repo so this problem was not detected.

PR #129, opened 2020-01-18
Documentations using Sphinx and readthedocs integration

ptpan:
This pull request sets up the documentation generation of PyMTL3. After this PR is merged, changes to documentations will be automatically picked up by readthedocs and update the PyMTL  3 documentation [here](https://pymtl3.readthedocs.io/).


docs/conf.py
```diff
@@ -0,0 +1,80 @@
...
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project   = 'PyMTL3'
+copyright = f'2020-{datetime.datetime.utcnow().year}, Batten Research Group'
```
jsn1993: maybe 2017?

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
```
jsn1993: since we now know how to switch branches, do we still need Makefile?

ptpan: I'd say let's keep this makefile so that we can preview the build locally... It's also much easier to debug the doc build with the makefile.

.gitignore
```diff
@@ -7,6 +7,7 @@
 # still have it appear clean to git.
 
 build*/
+_build/
```
jsn1993, line 10: what is _build/ for?

setup.py
```diff
@@ -20,12 +20,10 @@ def get_version():
...
   assert sys.version_info[0] > 2, "Python 2 is no longer supported!"
 
-  result = "?"
-  with open("pymtl3/__init__.py") as f:
-    for line in f:
-      if line.startswith("__version__"):
-        _, result, _ = line.split('"')
-  return result
+  _d = {}
+  with open(path.join(path.dirname(__file__), "pymtl3/version.py")) as f:
```
jsn1993, line 24: Do other projects also use version.py?

.gitignore
ptpan: Sphinx build directory

setup.py
ptpan: Hypothesis and mypy both use `version.py`

.gitignore
jsn1993: can it use build too?

ptpan: done

PR #131, opened 2020-01-22
Add ExternalTrace ImportConfig

ptpan:
`ExternalTrace` config will bring the Verilog line trace string to PyMTL through DPI call. It essentially ports the same functionality from PyMTL v2 to PyMTL 3.

### DIFF
--- a/pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
+++ b/pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
@@ -79,16 +79,16 @@ def test_reg_external_trace( do_test ):
 
   a.in_ = Bits32(1)
   a.tick()
-  assert a.line_trace().replace(' ', '') == 'q=0'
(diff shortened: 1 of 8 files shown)


pymtl3/passes/backends/sverilog/import_/test/ImportedObject_test.py
```diff
@@ -55,6 +56,40 @@ def construct( s ):
...
+  a = VRegTrace()
+  a.elaborate()
+  ipass = ImportPass()
+  a.config_sverilog_import.fill_missing( a )
+  a = ipass.get_imported_object( a )
+  a.apply( SimulationPass() )
+
+  a.in_ = Bits32(1)
+  a.tick()
+  assert a.line_trace().replace(' ', '') == 'q=0'
```
jsn1993: is it possible to check equality with space?

ptpan: Yes

PR #133, opened 2020-01-24
Fixed a few bugs, add run_sim and run_test_vector_sim

jsn1993:
I'm adding one early type check in update blocks. Basically I'm disallowing writes to components, interfaces, and method ports in update blocks because they don't make sense. We can still perform reads like ```isinstance( s.x, xxx )``` which might be helpful.

### DIFF
--- a/pymtl3/dsl/NamedObject.py
+++ b/pymtl3/dsl/NamedObject.py
@@ -21,7 +21,7 @@ Date   : Nov 3, 2018
 import re
 from collections import deque
 
-from .errors import NotElaboratedError
+from .errors import NotElaboratedError, PyMTLDeprecationError
 
 
 class DSLMetadata:
@@ -152,8 +152,13 @@ class NamedObject:
       if isinstance( obj, NamedObject ):
         ud = obj._dsl
 
+        try:
+          ud.level      = sd.level + 1
+        except AttributeError:
+          raise PyMTLDeprecationError("\n- Some component in the hierarchy implements __init__ "
+                                      "instead of construct(). Please FIX them!")
+
         ud.parent_obj = s
-        ud.level      = sd.level + 1
 
         ud._my_name  = ud.my_name = name
         ud.full_name = f"{sd.full_name}.{name}"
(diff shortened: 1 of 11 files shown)


pymtl3/stdlib/test/test_utils.py
```diff
@@ -67,3 +80,190 @@ def run_test( self ):
...
+  # Reset model
+
+  model.sim_reset()
+  print()
+
+  # Run simulation
+
+  ncycles = 0
+  while not model.done() and ncycles < max_cycles:
+    print( f"{ncycles:3}: {model.line_trace()}" )
```
yo96: i think we should add line trace as a option too. Also maybe we can let the user select which backend to use? instead of `test_verilog` we can make the argument `translation_backend` and default to None

pymtl3/dsl/NamedObject.py
```diff
@@ -152,8 +152,13 @@ def __setattr_for_elaborate__( s, name, obj ):
       if isinstance( obj, NamedObject ):
         ud = obj._dsl
 
+        try:
+          ud.level      = sd.level + 1
+        except AttributeError:
+          raise PyMTLDeprecationError("\n- Some component in the hierarchy implements __init__ "
```
ptpan: To be clear, is it true that any legal PyMTL component that could ever be is NOT allowed to override `__init__`? Is there any possible use of that?

jsn1993: hmmm maybe I can remove this

pymtl3/stdlib/test/test_utils.py
jsn1993: hmmm

jsn1993: ok, I can add the line trace function, but the test_verilog has a different meaning here ...

jsn1993: Basically test_verilog is a string option that selects verilator initialization values

yo96: i see, maybe another argument to select the backend then?

jsn1993: not important now

PR #134, opened 2020-01-24
Fix Verilog import translation

ptpan:


jsn1993:
is the plan to use the old caching mechanism for now?

PR #136, opened 2020-01-25
Generic method ifcs 

jsn1993:
I want to merge this in before the hypothesis branch, so I pulled out all the changes and make a PR.
several things in this PR:
1. generic method ifc FL/CL/RTL.
2. RTL will have ret for return value instead of mixing the use of msg everywhere
3. Enhance OpenLoopCL pass to support loops in RTL


pymtl3/stdlib/ifcs/SendRecvIfc.py
```diff
@@ -71,33 +81,16 @@ def connect( s, other, parent ):
...
-      trace_len = s.trace_len
-    except AttributeError:
-      s.trace_len = len( "{}".format( s.MsgType() ) )
-      trace_len = s.trace_len
-
-    return enrdy_to_str( s.msg, s.en, s.rdy, trace_len )
-
-  def __str__( s ):
-    return s.line_trace()
+    super().construct( en=True, rdy=True, MsgType=Type, RetType=None )
```
yo96: Our send recv interface will use `ret` instead of `msg` now? This will require us to change almost all our code... Besides I'm not sure if it is the right thing to do. If some RTL developer is not familiar with method based and just want to write RTL this naming convention can be quite confusing

pymtl3/dsl/Connectable.py
```diff
@@ -449,7 +482,62 @@ def __str__( s ):
...
+      s.rdy = OutPort( Bits1 )
+
+    if MsgType is not None:
+      s.msg = InPort ( MsgType )
+      s.MsgType = MsgType
+
+    if RetType is not None:
+      s.ret = OutPort( RetType )
+      s.RetType = RetType
+
```
yo96: Can we add some conditions here so that if MsgType is None and RetType is not None then s.msg is the OutPort? This way we don't have to change all of our code that's using send/recv

yo96: and maybe add a check to make sure at least one port is not None?

pymtl3/stdlib/ifcs/SendRecvIfc.py
yo96: Well sorry I mean enq/deq

pymtl3/stdlib/ifcs/GetGiveIfc.py
```diff
@@ -10,79 +10,97 @@
...
-
-    return enrdy_to_str( s.msg, s.en, s.rdy, trace_len )
-
-  def __str__( s ):
-    return s.line_trace()
+    super().construct( en=True, rdy=True, MsgType=None, RetType=Type )
 
 #-------------------------------------------------------------------------
 # GiveIfcRTL
 #-------------------------------------------------------------------------
```
yo96: wrong header name

yo96:
all changes  look good to me but I'm not sure if we should use `ret` instead of `msg` for our normal RTL Enq/Deq (Get/Give) interfaces... Enq will have .msg and Deq will have .ret. That makes sense in the method-based perspective but is less natural if think of it as dataflow. Besides that will require us to change a log of code. Maybe @cbatten can chime in here a bit to discuss about it and make a decision?

pymtl3/stdlib/ifcs/SendRecvIfc.py
jsn1993: Send/Recv will still use message. Enq will still use msg because I think directly moving to arg is too aggressive, but using ret for Deq(Give) is what we are doing as the first step. This is to avoid mixing the fields when someone tries to connect two very unrelated interfaces

pymtl3/dsl/Connectable.py
jsn1993: send/recv is unaffected. Only give/deq is affected

cbatten:
I am ok with trying the new naming scheme. As @jsn1993 said, for send/recv it will still be msg and for enq it will be msg too ... but for deq it will be ret ... which seems like the right thing to do because the ret value goes in the opposite direction as the enable. deq is really very much a method based interface so I think it is okay ... I like keeping msg because then for send/recv it kind of still looks like what we had before.

PR #137, opened 2020-01-28
add print_line_trace, refactor sim_reset

jsn1993:


pymtl3/dsl/test/sim_utils.py
```diff
@@ -497,3 +498,14 @@ def cleanup_signals( m ):
             cleanup_signals( obj )
 
   cleanup_signals( s )
+
+  def create_reset( top ):
+    def reset():
+      print()
```
yo96: why do we need a print here?

pymtl3/passes/sim/AddSimUtilFuncsPass.py
```diff
@@ -0,0 +1,50 @@
...
+
+    top.sim_reset = self.create_reset( top, reset_value=1 )
+    top.print_line_trace = self.create_print_line_trace( top )
+
+  @staticmethod
+  # Simulation related APIs
+  def create_reset( top, reset_value ):
+    assert reset_value in [0,1]
+    def reset(print_line_trace=False):
+      print()
```
yo96: `reset_value` doesn't sound right... It's like the default value that gets loaded into register when reset is enablesd. Instead I think it should be `active_high=True`. In this way we can get rid of that assertion as well.

Some unimportant stuff: empty space after `(` and before `)` and we probably can remove this print?

pymtl3/passes/sim/AddSimUtilFuncsPass.py
```diff
@@ -0,0 +1,50 @@
...
+      top.reset = b1( reset_value )
+      top.tick() # Tick twice to propagate the reset signal
+      if print_line_trace:
+        print( f"{top.simulated_cycles:3}r {top.line_trace()}" )
```
yo96: is `simulate_cycles` a first class thing? Or it has to be added by prior passes?

pymtl3/passes/sim/AddSimUtilFuncsPass.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+from pymtl3.datatypes import b1
+from pymtl3.passes.BasePass import BasePass, PassMetadata
+from pymtl3.passes.errors import PassOrderError
+
+
+class AddSimUtilFuncsPass( BasePass ):
+
+  def __call__( self, top ):
```
yo96: Can we have some arguments here like `active_high=True`?

pymtl3/dsl/test/sim_utils.py
jsn1993: We need this print for the first line in pytest

pymtl3/passes/sim/AddSimUtilFuncsPass.py
jsn1993: has to be added by the tick pass

pymtl3/dsl/test/sim_utils.py
yo96: I think it is fine to add a print() in each test case. If reset is called again during the simulation this will also print an empty line.

pymtl3/passes/sim/AddSimUtilFuncsPass.py
jsn1993: done

pymtl3/dsl/test/sim_utils.py
```diff
@@ -497,3 +498,13 @@ def cleanup_signals( m ):
             cleanup_signals( obj )
 
   cleanup_signals( s )
+
+  def create_reset( top ):
+    def reset():
+      top.reset = Bits1(1)
+      top.tick()
+      top.tick()
```
yo96: Can you remind me why we need to tick twice here? I remember previously it has something to do with top level ports. Do we still need to it now?

jsn1993: Ideally I want to create special tick functions for this ... because we haven't solved the problem you are talking about. Basically you need to call eval_combinational between the top.reset and top.tick, but I don't create eval_combinational if I detect any CL primitive.

PR #138, opened 2020-01-29
fix lambda block name when the signal has bracket, add translation tests

jsn1993:


yo96:
I just ran into this bug...

jsn1993:
> I just ran into this bug...

ha! it's already merged. Was the bug related to pymtl3-net?

yo96:
yes i was trying to translate a switch unit

PR #142, opened 2020-02-09
adding ascii mode for printwave_pass

kaishuocheng:
I add an option for ascii mode, although default mode is still text_fancy.


cbatten:
I think maybe the default should be ascii so things "just work" on repl.it?

kaishuocheng:
> I think maybe the default should be ascii so things "just work" on repl.it?

Yeap I can flip it. But there is a checker that forces the user to set the "._tracing" to be either "text_ascii" or "text_fancy". I think it is helpful so I leave it there, or maybe we want to remove that?

PR #145, opened 2020-02-21
Improve import pass

jsn1993:
I remove the connections and mangled_ports in sv import pass. As a by-product, I was actually able to drastically simplify yosys import pass. Now it overrides only the "mangling scheme" part of the general SV import pass .


pymtl3/passes/backends/sverilog/import_/ImportPass.py
```diff
@@ -831,122 +694,169 @@ def _get_arg_str( name, obj ):
...
+      dtype_nbits = dtype.get_length()
+
+      # If the top-level signal is a Bits, we directly
+
+      if isinstance( dtype, rdt.Vector ):
+        return s._gen_ref_read( lhs, rhs, dtype_nbits )
+
+      # If the top-level signal is a struct, we add the datatype to symbol?
+
+      if isinstance( dtype, rdt.Struct ):
```
ptpan: If `dtype` is a nested struct, does this work? I would imagine there would be a recursive process to breakdown the data type into bits before passing into `_gen_ref_read`?

jsn1993: hmm good point. Can you add a test to catch that?

jsn1993: This doesn't matter, because tmp is going to hold all the value. Line 835 will recursively break tmp down to the actualy nested struct. See line 836 where the final position is dtype,get_length.

PR #146, opened 2020-02-21
port my import pass improvements to 5745 branch

jsn1993:


jsn1993:
> Most of the changes are similar to the PR that was merged right? What is the status of the labs?

Yes, it is, but I also reimplement the unified gen_mapped_ports which selectively unpack port arrays when they are mapped ..

PR #152, opened 2020-02-26
connect_bits2bitstruct

yo96:
Very small PR that adds `connect_bits2bitstruct` in stdlib and a `get_type` API for connectables.

### DIFF
--- a/pymtl3/stdlib/connects/connect_bits2bitstruct.py
+++ b/pymtl3/stdlib/connects/connect_bits2bitstruct.py
@@ -22,7 +22,7 @@ def _connect_bits2bitstruct_h( field, bits_signal, slices ):
   # are very few use cases; 2) the pack order of a list of fields is
   # somewhat ambiguous.
   if isinstance( field, list ):
-    assert False, "A list of fields is not supported at the moment."
+    assert False, 'A list of fields is not supported at the moment.'
 
   field_type = field.get_type()
 
@@ -51,13 +51,19 @@ def connect_bits2bitstruct( signal1, signal2 ):
   type2 = signal2.get_type()
 
   if is_bitstruct_class( type1 ):
-    assert issubclass( type2, Bits )
+    assert issubclass( type2, Bits ), \
+      f'{type1.__qualname__} is a bitstruct ' \
+      f'but {type2.__qualname__ is not a Bits type!}'
     bits_signal, bitstruct_signal = signal2, signal1
+
   elif is_bitstruct_class( type2 ):
-    assert issubclass( type1, Bits )
+    assert issubclass( type1, Bits ), \
+      f'{type2.__qualname__} is a bitstruct ' \
+      f'but {type1.__qualname__ is not a Bits type!}'
     bits_signal, bitstruct_signal = signal1, signal2
+
   else:
-    assert False, "Can only connect a bitstruct wire to bits wire"
+    assert False, 'Can only connect a bitstruct wire to bits wire'
 
   # Connect field to corresponding slice
   _connect_bits2bitstruct_h( bitstruct_signal, bits_signal, [] )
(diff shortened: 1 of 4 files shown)


jsn1993:
@ptpan can this actually help simplifyactuallysimplify verilog import?

pymtl3/stdlib/connects/connect_bits2bitstruct.py
```diff
@@ -0,0 +1,63 @@
...
+  # somewhat ambiguous.
+  if isinstance( field, list ):
+    assert False, "A list of fields is not supported at the moment."
+
+  field_type = field.get_type()
+
+  if issubclass( field_type, Bits ):
+    start     = 0 if not slices else slices[-1].stop
+    stop      = start + field_type.nbits
+    cur_slice = slice( start, stop )
```
jsn1993: do you actually need to carry a list of slices? I guess you only need to record the last slice stop?

pymtl3/stdlib/connects/connect_bits2bitstruct.py
```diff
@@ -0,0 +1,63 @@
...
+    slices.append( cur_slice )
+    connect( field, bits_signal[ cur_slice ] )
+
+  else:
+    assert is_bitstruct_class( field_type )
+    fields_dict = getattr( field_type, _FIELDS )
+    fields_lst  = list( fields_dict.items() )
+    fields_lst.reverse() # reverse is faster than [::-1]
```
jsn1993: For name in reversed(list(items)) and .keys

pymtl3/stdlib/connects/connect_bits2bitstruct.py
```diff
@@ -0,0 +1,63 @@
...
+
+#-------------------------------------------------------------------------
+# connect_bits2bitstruct
+#-------------------------------------------------------------------------
+
+def connect_bits2bitstruct( signal1, signal2 ):
+  type1 = signal1.get_type()
+  type2 = signal2.get_type()
+
+  if is_bitstruct_class( type1 ):
```
jsn1993: Also add the error message in the two above assertions

pymtl3/stdlib/connects/connect_bits2bitstruct.py
yo96: Ideally I only need one integer, but after a recursive call changes the integer value it needs to be observable by the parent call. Originally I use a single element list to do this but it turns out to be a little confusing. If it is C++ I would just record an integer pointer...

yo96:
Another issue I'd like to discuss is whether we want to check the bitwidth before we do the connect? If so that means we need to call `get_nbits` on the bitstruct and as a result we traverse the tree twice. I was wondering if there is a way to avoid the repetitive tree traversal.

yo96:
> @jsn1993 Another issue I'd like to discuss is whether we want to check the bitwidth before we do the connect? If so that means we need to call `get_nbits` on the bitstruct and as a result we traverse the tree twice. I was wondering if there is a way to avoid the repetitive tree traversal.

whatever. I think I'm over optimizing things.

pymtl3/stdlib/connects/connect_bits2bitstruct.py
```diff
@@ -0,0 +1,75 @@
...
+    assert False, 'Can only connect a bitstruct wire to bits wire'
+
+  bw1 = get_nbits( type1 )
+  bw2 = get_nbits( type2 )
+  assert bw1 == bw2, 'Bitwidth mismatch! ' \
+    f'{type1.__qualname__} is {bw1}-bit ' \
+    f'but {type2.__qualname__} is {bw2}-bit.'
+
+  # Connect field to corresponding slice
+  _connect_bits2bitstruct_h( bitstruct_signal, bits_signal, 0 )
```
jsn1993: wanna add an assertion that checks if the 'end' value returned is the same as bw1?

PR #155, opened 2020-03-03
Major syntax changes

jsn1993:
This PR implements four important syntax changes:

- `__call__` connection is deprecated. This is somewhat convenient but the keyword argument approach doesn't work for cases when we need to connect individual ports in the interface.
- Signal now only takes Bits or bitstruct type. We used to allow arbitrary type but throughout the practice we don't find creating Wire(int) very useful.
- Signal now allows taking a single integer N as a shortcut for mk_bits(N). This might avoid a lot of mk_bits calls ... 
- Support `@update` and @update_ff. Dropping @s.update.

### DIFF
--- a/examples/ex03_proc/ProcRTL.py
+++ b/examples/ex03_proc/ProcRTL.py
@@ -53,112 +53,110 @@ class ProcRTL( Component ):
 
     # Bypass queues
 
-    s.imemreq_q   = BypassQueue2RTL( req_class, 2 )( deq = s.imem.req )
+    s.imemreq_q   = BypassQueue2RTL( req_class, 2 )
 
     # We have to turn input receive interface into get interface
 
-    s.imemresp_q  = BypassQueueRTL( resp_class, 1 )( enq = s.imem.resp )
-    s.dmemresp_q  = BypassQueueRTL( resp_class, 1 )( enq = s.dmem.resp )
-    s.mngr2proc_q = BypassQueueRTL( Bits32, 1 )( enq = s.mngr2proc )
-    s.xcelresp_q  = BypassQueueRTL( xresp_class, 1 )( enq = s.xcel.resp )
+    s.imemresp_q  = BypassQueueRTL( resp_class, 1 )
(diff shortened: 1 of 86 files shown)


examples/ex03_proc/ProcRTL.py
```diff
@@ -53,112 +53,110 @@ def construct( s ):
...
-      imemreq_addr  = s.imemreq_q.enq.msg.addr,
-      imemresp_data = s.imemresp_drop.out.ret,
+    s.dpath = m = ProcDpath()
 
-      # dmem ports
-      dmemreq_addr  = s.dmem.req.msg.addr,
-      dmemreq_data  = s.dmem.req.msg.data,
-      dmemresp_data = s.dmemresp_q.deq.ret.data,
+    # imem ports
+    m.imemreq_addr  //= s.imemreq_q.enq.msg.addr
```
ptpan: I imagine there could be a `stdlib.inspect_connect`  function that inspects two components and connect their ports based on the names... would be even better than SV `.*`

pymtl3/stdlib/rtl/enrdy_queues_test.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 
 from pymtl3 import *
-from pymtl3.stdlib.test import TestSinkCL, TestSrcCL
+from pymtl3.stdlib.test import TestSinkCL, TestSrcCL, run_sim
```
ptpan: Is the idea to prefer using `stdlib.run_sim` over customized ticking method? That sounds like a huge improvement

jsn1993: I think the problem is we had too many run_sim with basically the same code ... Now I just remove the redundant ones. I already did that in pymtl3-net.

examples/ex03_proc/ProcRTL.py
jsn1993: connect_auto in pymtl2 basically ..

PR #159, opened 2020-03-06
Verilog translation bug fix

ptpan:
1. Fixed a bug which occurs when a free variable is used as the step of for loop
2. Improved the translated BitStruct file info
3. Added a check which prevents BitStruct fields from using the names of BitStructs

### DIFF
--- a/pymtl3/passes/backends/verilog/translation/behavioral/VBehavioralTranslatorL2.py
+++ b/pymtl3/passes/backends/verilog/translation/behavioral/VBehavioralTranslatorL2.py
@@ -130,11 +130,12 @@ class BehavioralRTLIRToVVisitorL2( BehavioralRTLIRToVVisitorL1 ):
 
     begin    = ' begin' if len( node.body ) > 1 else ''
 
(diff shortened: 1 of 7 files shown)


pymtl3/passes/backends/verilog/translation/behavioral/VBehavioralTranslatorL2.py
```diff
@@ -130,11 +130,12 @@ def visit_For( s, node ):
 
     begin    = ' begin' if len( node.body ) > 1 else ''
 
-    cmp_op   = '<' if node.step.value > 0 else '<'
-    inc_op   = '+' if node.step.value > 0 else '-'
+    cmp_op   = '>' if node.step._value[31] else '<'
```
jsn1993: how does this _value[31] work? inferred as Bits32?

ptpan: Yes -- checking if the number is negative by looking at the MSB

PR #164, opened 2020-03-16
Better constant extraction during RTLIR generation

ptpan:
This PR allows the use of complex expressions in translatable update blocks, as long as they evaluate to a constant object at translation time. The following Python objects are considered to be "constant":
1. BitsN and BitStruct
2. BitsN() and BitStruct()
3. Python int objects
4. PyMTL functions: sext, zext, concat, reduce_*


ptpan:
@yo96 This should allow you to use class attributes as if they are from an enum. They will be translated to the value they hold.

PR #167, opened 2020-03-22
Verilog TBgen

jsn1993:
This PR implements VerilogTBGenPass, a pass that generates a SystemVerilog test harness for co-simulation with SV simulators like VCS. This tool (need a name) combines our "test vector based simulation" and Synopsys VCAT.

### DIFF
--- a/pymtl3/passes/backends/verilog/tbgen/test/VerilogTBGen_test.py
+++ b/pymtl3/passes/backends/verilog/tbgen/test/VerilogTBGen_test.py
@@ -23,6 +23,7 @@ from pymtl3.passes.backends.verilog import (
 from pymtl3.passes.rtlir.util.test_utility import do_test
 from pymtl3.stdlib.test import TestVectorSimulator
 
+from ...testcases import CaseConnectArrayBits32FooIfcComp
 
 def local_do_test( _m ):
   _m.elaborate()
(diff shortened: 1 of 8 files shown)


pymtl3/passes/backends/verilog/tbgen/test/VerilogTBGen_test.py
```diff
@@ -0,0 +1,104 @@
...
+def local_do_test( _m ):
+  _m.elaborate()
+  _m.apply( VerilogPlaceholderPass() )
+  m = TranslationImportPass()( _m )
+  m.verilog_tbgen = True
+  m.apply( VerilogTBGenPass() )
+  sim = TestVectorSimulator( m, _m._test_vectors, _m._tv_in, _m._tv_out )
+  sim.run_test()
+
+def test_normal_queue( do_test ):
```
ptpan: To be clear, the tests here just generate the testbench but do not simulate it?

pymtl3/passes/backends/verilog/tbgen/verilog_tbgen.v.template
```diff
@@ -0,0 +1,62 @@
+`define HOLD_TIME 1
```
ptpan: Does it make sense to have customized `HOLD_TIME` and `SETUP_TIME`? I'm not sure if anyone might need that but sounds like some nice options to have for the pass.

ptpan: For future work tho.. i think this should be good enough for now

pymtl3/passes/backends/verilog/tbgen/test/VerilogTBGen_test.py
jsn1993: i assume no verilog simulator ...

pymtl3/passes/backends/verilog/tbgen/verilog_tbgen.v.template
jsn1993: I think we will have to add a config for this pass. There are other parameters like different TB names for different test cases.

PR #170, opened 2020-03-25
@= 

jsn1993:


pymtl3/dsl/test/sim_utils.py
```diff
@@ -75,7 +75,7 @@ def simple_sim_pass( s, seed=0xdeadbeef ):
                         .replace( "(", "_" ).replace( ")", "" ) \
                         .replace( ",", "_" )
 
-        rstrs    = [ f"{x!r} = _w" for x in readers ]
+        rstrs    = [ f"{x!r} @= _w" for x in readers ] # THIS IS SLOW, NOW WE CAN HAVE BETTER MECHANISM
```
ptpan: Dumb question: what does the format string `{x!r}` do?

pymtl3/passes/autotick/OpenLoopCLPass.py
```diff
@@ -31,16 +32,19 @@
...
   def __init__( self, print_line_trace=True ):
-    self.line_trace_on = print_line_trace
+    self.print_line_trace = print_line_trace
 
   def __call__( self, top ):
     if not hasattr( top._dag, "all_constraints" ):
       raise PassOrderError( "all_constraints" )
 
     top._sched = PassMetadata()
+    top._sim   = PassMetadata()
```
ptpan: I think before the release we are gonna use the metadata APIs and deprecate `PassMetadata`?

pymtl3/passes/backends/verilog/util/test_utility.py
```diff
@@ -290,7 +290,7 @@ def outport_filter( obj ):
   def ref_tv_out( model, test_vector ):
     dct = {}
     for out_port in all_output_ports:
-      dct[ out_port ] = eval( "model." + out_port._dsl.my_name )
+      dct[ out_port ] = eval( "model." + out_port._dsl.my_name ).clone() # WE NEED TO CLONE NOW
```
ptpan: Is this because we always reuse and mutate the same Bits object now?

jsn1993: right. I'm adding another function to catch = instead of @= at the top level.

pymtl3/dsl/test/sim_utils.py
jsn1993: repr(x)

pymtl3/passes/autotick/OpenLoopCLPass.py
jsn1993: You are correct. I'm not 100% sure about how to use metadata APIs now and those code are developed before the metadata APIs are merged. Can you point me to an example?

PR #174, opened 2020-03-31
Fixed amo method in MemIfcFL2CLAdapter

qtt2:
There was a missing data field in requests made in `amo` method of `MemIfcFL2CLAdapter`.


jsn1993:
I never really used the AMO there ... thanks for fixing this

qtt2:
Should I go ahead and merge it even though the codecov/path is not successful? I guess we don't have tests for this?

ptpan:
> Should I go ahead and merge it even though the codecov/path is not successful? I guess we don't have tests for this?

Right -- currently we do not enforce the code coverage check...

PR #178, opened 2020-04-01
Pytest plugin and a few import improvements

ptpan:
1. Implemented `pytest-pymtl3` pytest plugin. This plugin provides the commonly used `dump_vcd` and `test_verilog` fixtures.
2. During translation and pickling, if a component does not have parameters, its name will be `<component_name>_noparam`.
3. During translation and pickling, the generated files now always have a `__pickled` suffix.
4. Assigning BitStructs to Bits and Bits to BitStructs are allowed.
5. `port_map` is now a dict from port objects to their names.


ptpan:
I found that the `pytest-cov` plugin which generates the coverage report has to be the one that performs the initial import of the `pymtl3` module. If we put the `pytest-pymtl3` plugin under the `pymtl3` namespace, `pytest-cov` will not be able to generate coverage. This is annoying and as a temporary work-around I have to put our plugin under the top level `plugins` folder...

.travis.yml
```diff
@@ -37,7 +37,7 @@ install:
  # Install Python requirements
  - pip install --upgrade pip setuptools twine
  - pip install --requirement requirements/CI.txt
- - pip install .
+ - pip install --editable .
```
jsn1993, line 40: i'm still not sure on this editable thing. What is the convention of doing this? Shipping tests might not be that bad? Are the tests changing locations when they are installed without editable?

examples/ex02_cksum/test/ChecksumRTL_test.py
```diff
@@ -106,20 +109,20 @@ def setup_class( cls ):
...
   # See pytest documetnation for more details.
   def teardown_method( s, method ):
-    s.vcd_file_name = ""
+    s.current_test_method_name = ""
 
   def run_sim( s, th, max_cycles=1000 ):
 
+    s.vcd_file_name = ""
+
+    if s.__class__.cmdline_opts["dump_vcd"]:
```
jsn1993, line 123: why is it s.__class__?

jsn1993: it seems radically different from the old one and I'm not sure how the new one works.

examples/ex03_proc/test/ProcFL_test.py
```diff
@@ -156,11 +156,11 @@ def test_srl_rand_delays( s, dump_vcd ):
     asm_test( inst_bne.gen_value_test             ),
     asm_test( inst_bne.gen_random_test            ),
   ])
-  def test_bne( s, name, test, dump_vcd ):
+  def test_bne( s, name, test, cmdline_opts ):
```
jsn1993, line 159: why do you need cmdline_opts in FL tests?

.travis.yml
ptpan: Without `--editable`, `pip` seems to copy the files (including the tests) into `site-packages`. When pytest tries to collect the tests, it will see two different versions of test files: ones in the git repo and ones under `site-packages`. `--editable` solves this issue and I think maybe that should be what we recommend to pymtl3 developers?

examples/ex02_cksum/test/ChecksumRTL_test.py
ptpan: See this line: https://github.com/cornell-brg/pymtl3/blob/846187d57d32a4fa9537ff6afe8dd136d8751a1c/pytest_plugin/pytest_pymtl3.py#L43
That line dynamically injects `cmdline_opts` as an attribute of the test class. We can also inject that to test functions but since most test functions will have `cmdline_opts` in their arguments that might not be necessary.

examples/ex03_proc/test/ProcFL_test.py
ptpan: because `dump_vcd` is now part of the unified `cmdline_opts`? also i think we initially had this in FL because we want to reuse tests across multiple levels?

.travis.yml
ptpan: * `--editable` solves this issue by _not_ copying the files and leaving all source code in the git repo.

examples/ex03_proc/test/ProcFL_test.py
jsn1993: hmmm right, we don't really need dump_vcd for FL model but yeah, reusing tests...

.travis.yml
jsn1993: site-packages is in the virtual environment, right? if you do pytest .. it shouldn't collect tests out of virtual environment

ptpan: It does. [This build](https://travis-ci.com/github/cornell-brg/pymtl3/jobs/311704515) failed because of that and you can use a debug build to see exactly what happened. I can't remember the exact error message but it said two different modules for the test same were found (one under `site-packages` and one in the repo).

jsn1993: Why didn't we see this failure before?

examples/ex02_cksum/test/ChecksumRTL_test.py
ptpan: they are the same?
https://github.com/cornell-brg/pymtl3/blob/301fec576e327114468ec1fb81b31c57303b6d6b/conftest.py#L28

ptpan: `cmdline_opts['dump_vcd']` has the same value as the previous `dump_vcd` fixture?

jsn1993: the fixture dump_vcd actually comes up with the string, but here the run_sim function assembles the string

ptpan: I think Yanghui did this because he was not able to get the fixture value. I'll change this since now we can use the fixture.

PR #179, opened 2020-04-01
refactor gen signal expr

jsn1993:
simplify logic


jsn1993:
@ptpan take a quick look?

PR #181, opened 2020-04-12
GC fix & translation performance opt

jsn1993:
I wanted to fix the gc failure that caused out of memory when running a large amount of big tests, when a module is translated and then imported. Then I found that the _rtlir_cache is a global object which actually contains pointers to the components. I had to reimplement the RTLIR_cache.

Meanwhile, I did some profiling and managed to get rid of lots of overheads during translation.


pymtl3/passes/backends/verilog/test/TranslationImport_adhoc_gc_test.py
```diff
@@ -0,0 +1,97 @@
...
+    _tv_out_str += '  pass'
+  for attr, Bits, idx in zip( args[0::3], args[1::3], args[2::3] ):
+    if isinstance(Bits, str):
+      _tv_out_str += f'  assert m.{attr} == {Bits}( tv[{idx}] )\n'
+    else:
+      _tv_out_str += f'  assert m.{attr} == {Bits.__name__}( tv[{idx}] )\n'
+  exec( _tv_out_str, globals(), local_dict )
+  return local_dict['tv_out']
+
+class CasePlusOneComp:
```
ptpan, line 51: Is there a reason why this test case and `_set``_check` are not in `verilog/testcases/`?

jsn1993: oh I totally forgot about this test. I will refactor the code

jsn1993: fixed

PR #185, opened 2020-05-17
stdlib reorg + passes reorg

jsn1993:
This PR is probably the last major revamp of the whole software hierarchy before we make the official release
- stdlib has a new organization
- using verilog backend should invoke "from pymtl3.passes.backends.verilog import *"


examples/ex03_proc/test/harness.py
```diff
@@ -14,9 +14,8 @@
 from examples.ex03_proc.NullXcel import NullXcelRTL
 from examples.ex03_proc.tinyrv0_encoding import assemble
 from pymtl3 import *
-from pymtl3.stdlib.cl.MemoryCL import MemoryCL
+from pymtl3.stdlib.mem.MagicMemoryCL import MagicMemoryCL, mk_mem_msg
```
ptpan, line 17: People can still do `from pymtl3.stdlib.mem import MagicMemoryCL` right?

pymtl3/passes/backends/verilog/tbgen/VerilogTBGenPass.py
```diff
@@ -21,19 +22,28 @@
...
+  case_name = MetadataKey(str)
+
   def __call__( self, top ):
     if not top._dsl.constructed:
       raise VerilogImportError( top,
         f"please elaborate design {top} before applying the TBGen pass!" )
 
-    top._tbgen = PassMetadata()
-    top._tbgen.tbgen_hooks = []
+    top._vtbgen = PassMetadata()
```
ptpan, line 39: Why not use the metadata APIs on `top` instead of having an attribute here? Ideally the way different passes communicate is through the metadata APIs?

pymtl3/passes/tracing/CLLineTracePass.py
```diff
@@ -12,10 +12,24 @@
...
   def __init__( self, default_trace_len=8 ):
     self.default_trace_len = default_trace_len
 
   def __call__( self, top ):
+
+    # Turn on by default
+    if top.has_metadata( self.enable ) and top.get_metadata( self.enable ) is False:
+      return
+
     if not hasattr( top, "_tracing" ):
```
ptpan, line 33: Similar here. Do we want to leave this as is for 3.0? I mean in general I don't think exposing internal data as an attribute of the component is a good idea. That makes sense only in cases where the pass users are meant to use the added attributes (like `sim_tick()`).

pymtl3/passes/backends/verilog/tbgen/VerilogTBGenPass.py
jsn1993: I forgot to do that. Do you have an example?

ptpan: For example, the import pass needs to know the file name of the translated result. The translation pass declares that as a metadata key here: https://github.com/cornell-brg/pymtl3/blob/b8debb763c46e96514642d1a4b45ea13c6f19522/pymtl3/passes/backends/verilog/translation/VerilogTranslationPass.py#L100
The ImportConfigs uses that metadata here: https://github.com/cornell-brg/pymtl3/blob/b8debb763c46e96514642d1a4b45ea13c6f19522/pymtl3/passes/backends/verilog/import_/VerilogVerilatorImportConfigs.py#L200

pymtl3/stdlib/queues/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .cl_queues import BypassQueueCL, NormalQueueCL, PipeQueueCL
+from .enq_deq_ifcs import DeqIfcRTL, EnqIfcRTL
+from .queues import BypassQueueRTL, NormalQueueRTL, PipeQueueRTL
+
+#  from .enrdy_queues import BypassQueue1RTL, NormalQueue1RTL, PipeQueue1RTL
```
ptpan, line 5: Is it safe to remove these if we are certain we'll never use val-rdy queues?

ptpan: Em.. seems like we still have the queues but we don't recommend using them and therefore choose not to expose them?

cbatten:
I think we should remove valrdy stuff for now including valrdy interface and valrdy queues ... we can maybe add them back in later if we really need to ... but since we are migrating everything to enrdy it seems to make sense to focus on that first? otherwise people might use valrdy queues when we would rather they not (I have already seen this happen)

pymtl3/passes/backends/verilog/tbgen/VerilogTBGenPass.py
jsn1993: I guess I need to modify every top._xxx added by those simulation passes and such

pymtl3/passes/tracing/CLLineTracePass.py
jsn1993: I fixed most of them. For simulation-related passes, we can fix them later

PR #186, opened 2020-05-19
official 3.0 release

jsn1993:


pymtl3/passes/autotick/OpenLoopCLPass.py
```diff
@@ -20,27 +20,29 @@
...
 class OpenLoopCLPass( BasePass ):
   def __init__( self, print_line_trace=True ):
-    self.line_trace_on = print_line_trace
+    self.print_line_trace = print_line_trace
 
   def __call__( self, top ):
     if not hasattr( top._dag, "all_constraints" ):
       raise PassOrderError( "all_constraints" )
 
     top._sched = PassMetadata()
```
ptpan, line 39: Just leave this as a TODO: i think our goal is still to get rid of the `PassMetadata` and turn to use metadata APIs? One possible way to do this is to have all passes that have to communicate through the `_sched` and `_sim` namespace inherit from a dummy base pass that defines the necessary metadata. See https://github.com/pymtl/pymtl3/blob/a3d5a512d57969e1f37f1eecbb0c8e438079c88b/pymtl3/passes/rtlir/behavioral/BehavioralRTLIRGenL1Pass.py#L45

pymtl3/passes/sim/PrepareSimPass.py
```diff
@@ -0,0 +1,298 @@
...
+  def __init__( self, print_line_trace=True, reset_active_high=True ):
+    assert reset_active_high in [ True, False ]
+
+    self.print_line_trace  = print_line_trace
+    self.reset_active_high = reset_active_high
+
+  def __call__( self, top ):
+    if hasattr(top, "sim_reset"):
+      raise AttributeError( "Please rename the attribute top.sim_reset")
+    if hasattr(top, "print_line_trace"):
```
ptpan, line 38: Have we decided to rename `line_trace` to `print_line_trace`?

pymtl3/passes/autotick/OpenLoopCLPass.py
jsn1993: Yes we will transition to APIs for sim and such. It's just that for _sched and _sim I haven't decided if we need to use a bass schedule pass or something else to store common metadata keys ...

pymtl3/stdlib/test/test_sinks.py
```diff
@@ -9,7 +9,7 @@
 """
 
 from pymtl3 import *
-from pymtl3.stdlib.ifcs import RecvIfcRTL, RecvRTL2SendCL, enrdy_to_str
+from pymtl3.stdlib.ifcs import RecvIfcRTL, RecvRTL2SendCL
```
ptpan, line 12: We talked about the plan of moving `stdlib.test` to `stdlib.test_utils` or whatever we feel is appropriate because `test` sounds like it is the directory for test cases?

jsn1993: i forgot about that. Thanks for the reminder. Let me fix them.

pymtl3/passes/sim/PrepareSimPass.py
jsn1993: I haven't changed the line_trace convention. The print_line_trace function is a generated function that calls line_trace .

setup.py
```diff
@@ -77,9 +75,9 @@ def get_long_description():
 
   package_data={
     'pymtl3': [
-      'passes/backends/verilog/import_/verilator_wrapper.c.template',
-      'passes/backends/verilog/import_/verilator_wrapper.py.template',
-      'passes/backends/verilog/tbgen/verilog_tbgen.v.template',
+      # 'passes/backends/verilog/import_/verilator_wrapper.c.template',
```
ptpan, line 78: Now that these templates are just importable python objects we can get rid of `package_data` here...

PR #194, opened 2020-07-11
Implement add_value_port API and fix verilator bugs and fix formatting tools

jsn1993:


.travis.yml
```diff
@@ -48,7 +48,7 @@ install:
 script:
  - autoflake --recursive --in-place --remove-duplicate-keys .
  - pyupgrade --py3-only --keep-percent-format $(find . -name '*.py')
- - isort --apply --recursive ./pymtl3
+ - isort ./pymtl3
```
ptpan, line 51: Does isort perform a recursive formatting by default now?

pymtl3/passes/adhoc_transform/AddDebugSignalPass.py
```diff
@@ -0,0 +1,42 @@
...
+
+from pymtl3.dsl import Component, InPort, MetadataKey, OutPort
+from pymtl3.passes.BasePass import BasePass
+
+
+class AddDebugSignalPass( BasePass ):
+
+  debug_pins = MetadataKey(set)
+
+  def __call__( self, top, signal_names ):
```
ptpan, line 22: In this case you cannot `s.apply` the `AddDebugSignalPass` because it requires an extra argument to the `__call__` method... it would be nice if we can somehow make `apply` work here

pymtl3/passes/backends/verilog/import_/VerilogVerilatorImportConfigs.py
```diff
@@ -26,6 +26,12 @@ class VerilogVerilatorImportConfigs( BasePassConfigs ):
     # Enable verbose mode?
     "verbose" : False,
 
+    # Trade off compilation time for fast simulation performance?
+    # controls: vl_opt_level, vl_unroll_count
+    "fast": False,
```
ptpan, line 31: How about `optimized_c_compilation` or something similar?

pymtl3/dsl/Component.py
```diff
@@ -714,19 +723,113 @@ def replace_component_with_obj( top, foo, new_obj, check=True ):
...
+  def add_connection( top, o1, o2 ):
+
+    # Currently only support connecting signals
+
+    top._check_called_at_elaborate_top( "add_connection" )
+    if isinstance( o2, Connectable ): o1, o2 = o2, o1
+
+    assert isinstance( o1, Connectable ), "Cannot connect two non-connectable objects"
+
+    assert isinstance( o1, Signal ), "Currently only support connecting signals/constants"
```
ptpan, line 751: I wonder if it is possible (eventually) to connect a lambda block to a signal.. it's basically adding another concurrent process to the component and I'm sure designers will appreciate this post-elaboration flexibility.

.travis.yml
jsn1993:
```

5.0.0 Penny - July 4, 2020
Breaking changes:
...
--recursive option has been removed. Directories passed in are now automatically sorted recursive.
--apply option has been removed as it is the default behaviour.
...
```

pymtl3/dsl/Component.py
jsn1993: This is a very good point. The question is there is no "s" to catch. Basically we need the user to provide raw string as source code.

pymtl3/passes/adhoc_transform/AddDebugSignalPass.py
jsn1993: I agree. I didn't find a good solution

PR #199, opened 2020-08-24
Fix a placeholder caching issue

ptpan:
Fixed an issue where the translation result of a placeholder component disables import caching.


ptpan:
@jsn1993 I think a minor release including this PR and #198 would be nice.. what do you think?

PR #201, opened 2020-09-28
tbgen set values during reset

jsn1993:


ptpan:
I don't recall having reset issues with VTBGen? Closing this because of inactivity and feel free to reopen in case we notice any VTBGen reset problems.

PR #203, opened 2021-02-25
save pickled file as string

jsn1993:


ptpan (approved):
I like the change of making placeholder intermediate results transparent — build directories are much cleaner now. Please do remember to fix that pypy related CI issue...

PR #207, opened 2021-03-15
On-demand VCD dumping

ptpan:
This PR adds new metadata support to enable VCD dumping only for an ROI. In order to do this your component hierarchy needs a top level port (active-high) that enables VCD dumping.


cbatten:
This is exciting! Would love to get this also merged into the ECE 5745 branch. This would be _super_ helpful for teaching since we only want to evaluate the ROI. So for example if you look at this tutorial:

 - https://cornell-ece5745.github.io/ece5745-tut9-xcel

There is a whole thing where we have to estimate the energy overhead of initialization and verification by using an empty ROI and then subtract that out ... but this turns out to not be very accurate ... @jsn1993 can you take a look and see maybe if we can get this merged into the E
CE 5745 branch if we are happy with it?

jsn1993 (approved):

PR #216, opened 2021-07-02
Merge changes during ECE5745 2021

jsn1993:


jsn1993:
@ptpan There are a few V import failures that I tried but wasn't able to fix ... Can you please take a look?

pymtl3/passes/backends/verilog/VerilogPlaceholderPass.py
```diff
@@ -410,7 +410,13 @@ def _get_auto_prefix( s, m ):
     if 'placeholder' in pymtl_config and \
        'auto_prefix' in pymtl_config['placeholder'] and \
        pymtl_config.getboolean( 'placeholder', 'auto_prefix' ):
-      return f"{os.path.basename(parent_dir)}_"
+      if cwd == parent_dir:
```
ptpan, line 413: I updated the way the placeholder pass generates auto-prefix -- does this match your intentions?

pymtl3/passes/backends/verilog/test/TranslationImport_stdlib_test.py
```diff
@@ -124,7 +124,6 @@ def tv_out( model, tv ):
 # Swap run_test_queue with modified version that calls the local do_test
 def _run_queue_test( do_test, test_func ):
 
-  assert False
```
jsn1993: @ptpan Can you double check if the stdlib tests in /verilog uses the verilog backend for translation and yosys backend for yosys? The way it runs the test seems to be different now.

pymtl3/passes/backends/verilog/VerilogPlaceholderPass.py
jsn1993: I think yes. Why was it broken before you fixed it?

ptpan: If there's a `pymtl.ini` under `parent_dir` it would just return a single underscore -- the expected behavior would be to return the base name of `parent_dir`.

pymtl3/passes/backends/verilog/test/TranslationImport_stdlib_test.py
ptpan: Just updated that test case to work with the new queue testbench

pymtl3/passes/backends/yosys/test/TranslationImport_stdlib_test.py
```diff
@@ -10,17 +10,20 @@
 files and add my own test function (the `run_test`s). However we need
 to make sure the orignal reference is not lost and is restored after
 finishing each test (no matter it fails or passes).
+
+The following queue test cases are not ported to the Yosys backend
+because they don't use the test vector simulator:
+  test_normal1_simple,
```
jsn1993, line 16: @ptpan Is it possible to have _run_queue_test for yosys backend similar to what you did above?

ptpan: Done. Maybe there's a better way to avoid duplication of `_run_queue_test`...

PR #223, opened 2022-02-25
[stdlib] fix val/rdy autoconnect

yo96:
This PR includes the following changes:
 - Implemented CL/RTL auto-connection for val/rdy interfaces
 - Removed broken val/rdy master/minion interfaces


pymtl3/stdlib/stream/ifcs.py
```diff
@@ -24,11 +33,63 @@ def construct( s, Type ):
...
+  def connect( s, other, parent ):
+
+    # SendCL (other) -> [ RecvCL -> SendRTL ] -> RecvRTL
+    if isinstance( other, CallerIfcCL ):
+      connect_valid = True
+      # Create an adapter and add to parents
+      m = RecvCL2SendRTL( s.MsgType )
+
+      if hasattr( parent, "RecvCL2SendRTL_count" ):
+        count = parent.RecvRTL2SendCL_count
```
ptpan, line 48: This branch is not covered by our regression tests. Did you mean `RecvCL2SendRTL_count` instead of `RecvRTL2SendCL_count`?

pymtl3/stdlib/stream/ifcs.py
```diff
@@ -39,9 +100,58 @@ def construct( s, Type ):
...
     s.trace_len = len(str(Type()))
 
+  def connect( s, other, parent ):
+    # SendRTL -> [ RecvRTL -> SendCL ] -> RecvCL (other)
+    if isinstance( other, CalleeIfcCL ):
+      # Create an adapter and add to parents
+      m = RecvRTL2SendCL( s.MsgType )
+
+      if hasattr( parent, "RecvRTL2SendCL_count" ):
+        count = parent.RecvCL2SendRTL_count
```
ptpan, line 110: This line seems to be symmetric to the line above. I don't quite get why `RecvCL2SendRTL_count` is used here instead of `RecvRTL2SendCL_count`...

yo96:
Do we still want to keep the CL to RTL(val/rdy) auto-connection? If so then we may still need the fix.

ptpan:
Closing this due to inactivity

PR #224, opened 2022-03-03
Add a Resetable RegisterFile to stdlib basic rtl

jbrzozo24:
Needed for fixing ECE5745 blocking cache for use in 4 state simulation

### DIFF
--- a/pymtl3/stdlib/basic_rtl/__init__.py
+++ b/pymtl3/stdlib/basic_rtl/__init__.py
@@ -15,5 +15,5 @@ from .arithmetics import (
 )
 from .crossbars import Crossbar
 from .encoders import Encoder
-from .register_files import RegisterFile
+from .register_files import RegisterFile, RegisterFileRst
 from .registers import Reg, RegEn, RegEnRst, RegRst
--- a/pymtl3/stdlib/basic_rtl/register_files.py
+++ b/pymtl3/stdlib/basic_rtl/register_files.py
@@ -34,3 +34,46 @@ class RegisterFile( Component ):
         for i in range( wr_ports ):
           if s.wen[i]:
             s.regs[ s.waddr[i] ] <<= s.wdata[i]
+
+class RegisterFileRst( Component ):
+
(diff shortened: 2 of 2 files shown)


ptpan:
The implementation looks good to me. My concern is that `stdlib.basic_rtl` has always been a set of synthesizable components, and `RegisterFileRst` is not one of them. Maybe you can move your implementation of `RegisterFileRst` to a new file under `stdlib.mem` (not all memories in that directory are synthesizable)? And also add a comment to the class saying this is just modeling and not for synthesis?

cbatten:
But it is supposed to be for synthesis?

ptpan:
> But it is supposed to be for synthesis?

I mean you cannot easily swap in an SRAM bank for `RegisterFileRst` as you can do for `RegisterFile`. That's why `RegsiterFile` is under `basic_rtl` and also why I wanted to keep `RegisterFileRst` separate from `basic_rtl`.

cbatten:
You cannot swap a SRAM bank in for RegisterFile because a register file is combinational read ... I think RegisterFileRst should go in `basic_rtl` since it is a simple synthesizable component ... it is kind of completely separate from SRAMs?

ptpan:
> You cannot swap a SRAM bank in for RegisterFile because a register file is combinational read ... I think RegisterFileRst should go in `basic_rtl` since it is a simple synthesizable component ... it is kind of completely separate from SRAMs?

You are right.. I was thinking of register files implemented as SRAMs but rather `stdlib.basic_rtl.RegisterFile` is cleraly an array of registers.. Yeah it should go to basic_rtl then.

ptpan (approved):
Looks good to me

ptpan:
@jbrzozo24 As per our convention, you will be the person to hit the "merge" button once a PR is approved... I'll make another minor release after these commits go into master.

jbrzozo24:
I don't believe I have write access to this repo?

ptpan:
> I don't believe I have write access to this repo?

Oops.. I was too used to that convention when almost all PRs come from the core pymtl3 devlopement team. It's nice to have your contribution to the repo!

PR #231, opened 2022-09-12
get tests running on ProcFL

cbatten:
I fixed a few things to get all of the TinyRV0 tests running on ProcFL. I also switched things around so all of the tests are in ProcFL_test.py and then ProcRTL_test.py and ProcVRTL_test.py just inherit from ProcFL_test.py.


ptpan (approved):
Can confirm ProcFL/RTL/VRTL_test pass on brg server.

Issue #255, opened 2023-11-28
Driver with signals are shown as no drivers


KelvinChung2000:
The following example:

```
from pymtl3 import * 
from pymtl3.passes.backends.verilog import *


class a(Component):
    def construct(self):
        self.in_ = InPort(Bits32)
        self.out = OutPort(Bits32)

        @update_ff
        def test():
            self.out <<= self.in_

class b(Component):
    def construct(self):
        self.in_ = InPort(Bits32)
        self.out = OutPort(Bits32)

        @update_ff
        def test():
            self.out <<= self.in_


class top(Component):
    def construct(self):
        self.in_ = InPort(Bits32)

        self.a = a()
        self.b = b()
        
        self.a.in_ //= self.in_

        self.b.in_ //= self.a.out


if __name__ == "__main__":
    m = top()
    m.elaborate()
    m.set_metadata( VerilogTranslationPass.enable, True )
    m.apply(VerilogTranslationPass())
```

Yield the following error:

```
pymtl3.dsl.errors.NoWriterError: The following nets need drivers.
Net:
 - s.a.out
 - s.b.in_ 
```

This is not what I expected, as `a.out` is driven by `a.in_`  and `a.in_` is driven by `in_`.

ptpan:
Hi Kevin,

Thanks for filing this bug! I'm able to reproduce this locally and get the same error message as yours.

I also found that if you replace all `self` with `s` (e.g., `s.in_ = InPort(Bits32)`, `s.a.in_ //= s.in_`), this error goes away and produces the translation result. So it seems like the elaboration function assumes some the name of the current component needs to be `s`... I'm not exactly sure where that assumption happens though -- need to investigate further.

cbatten:
Thanks for the quick follow up @ptpan ! Odd that the elaboration pass makes an assumption the name of `self` vs `s` ... hopefully we can remote that restriction? @jsn1993 any thoughts on this?

KelvinChung2000:
After changing to `s`, the "supposed to have" errors pop up, like `@=` vs `=` in the `@update` block. Before I used `self`, they would just go through all the checks when I used `=`  in `@update` and generate some Verilog.

ptpan:
> After changing to `s`, the "supposed to have" errors pop up, like `@=` vs `=` in the `@update` block. Before I used `self`, they would just go through all the checks when I used `=` in `@update` and generate some Verilog.

[...]

PR #260, opened 2023-12-07
Add support for non-s args to refer to self

ptpan:
Addresses #255. This PR allows the use of positional arguments that is not "s" to refer to the self variable. An example of this is

```
class A(Component):
    def construct(self):
        self.in_ = InPort(Bits32)
        self.out = OutPort(Bits32)

        @update_ff
        def test():
            self.out <<= self.in_
```


cbatten:
Is this update generic meaning it will now work with any name? or is not hard coded to explicitly work with both `s` and `self`?

cbatten:
excellent!

PR #272, opened 2024-04-02
Adding __len__ interface and formatting to black

KelvinChung2000:
Adding a `__len__` interface to the `Bits` class. Finding the size of something with `len(bits)` is intuitive and easy to understand. 

I have formatted the code according to [black](https://github.com/psf/black) which makes it more pleasant to read and matches the appearance of modern Python code. If the original is still preferred, I will push it again without formatting it.


yo96:
Hi Kelvin,

Thank you so much for taking the time to submit this pull request! I really appreciate your contribution and the `__len__` interface you added to `Bits` looks good to me. 

Regarding the auto formatting, I understand where you're coming from and I agree that it can be beneficial in many cases. However, for this particular PR, if it's alright with you, I'd prefer to stick with the original coding style. We actually experimented with applying auto formatting using flake8 a while back, but ultimately decided to revert it. The reason being, we found that some of our manual formatting tricks, like aligning the `=` signs, really improved readability in certain scenarios, and we didn't want to lose that.

That being said, we definitely recognize the advantages of auto formatting, such as maintaining a consistent and canonical codebase. It's something we're actively considering and weighing the pros and cons of. If we do decide to move forward with auto formatting in the future, our plan would be to apply it to the entire codebase at once to ensure consistency throughout.

I hope you understand our perspective on this. Please let me know if you have any other thoughts or suggestions – I'm always happy to discuss further! Once again, thank you for your contribution!

pymtl3/datatypes/PythonBits.py
```diff
@@ -528,4 +531,4 @@ def oct( self ):
 
   def hex( self ):
     str = "{:x}".format(int(self._uint)).zfill(((self._nbits-1)//4)+1)
-    return "0x"+str
+    return "0x"+str
```
yo96, line 534: Can we preserve the new line at the end of the file here? Some linter or style guide checker may complain about this.

yo96 (commented):
Can you also add a test case for the new `len` interface in `pymtl3/datatypes/test/bits_test.py`?

yo96:
LGTM. Thanks for the PR!

yo96 (approved):

PR #277, opened 2024-05-31
fixing verilog placeholder pass

KelvinChung2000:
The placeholder pass wrapper does not add a clock and reset signal, even when the metadata is set,  which causes a missing pin warning from Verilator. 

Furthermore, the name of the clock and reset might vary between designs; two new parameters are added so that the user can define the name of the clock signal and the reset signal.


cbatten:
@ptpan can you take a look?

yo96:
Hmm.. As I recall we had support for this before. Not sure why it is missing?

KelvinChung2000:
I also saw that in the Doc, which surprised me, this is not working. When I do a repo-wide search, the `has_clk` and `has_reset` are never set to true anywhere. Maybe the feature forgot merging into main?

PR #278, opened 2024-06-03
Improving concat

KelvinChung2000:
Allowing concat to take list or generator to enable syntax like `s.a //= concat(b2(1) for _ in range(4))` to be used.

### DIFF
--- a/pymtl3/datatypes/helpers.py
+++ b/pymtl3/datatypes/helpers.py
@@ -9,6 +9,7 @@ Author : Shunning Jiang
 Date   : Nov 3, 2017
 """
 import math
+import types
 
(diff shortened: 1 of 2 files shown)


pymtl3/datatypes/helpers.py
```diff
@@ -18,7 +19,14 @@
...
     value = nbits = 0
 
+    if len(args) == 1:
+      if isinstance(args[0], list):
+        args = args[0]
+      elif isinstance(args[0], types.GeneratorType):
+        args = args[0]
+
     for x in args:
+      print(x)
```
yo96: We should remove the `print`

PR #289, opened 2025-06-16
Fix clog2 function to be fast and universal

GregoryMorse:
Using floor of log2 with IEEE double floats here, suffers slowness and inaccuracy at huge values where the log2 would round incorrectly due a 53-bit mantissa.


yo96:
Thank you for the PR! The change looks good to me.

cbatten:
@yo96 feel free to merge if you think it is good!

PR #291, opened 2025-07-07
Output VCD file compliant with the standard

### LINKED ISSUE #269: Incorrect VCD file format
I am trying to use a different waveform viewer ([surfer](https://gitlab.com/surfer-project/surfer)), which is not working. I have raised the issue to the `Surfer` team and found out there might be an encoding error in the VCD file for pymtl. 

This is the link to the [issue](https://gitlab.com/surfer-project/surfer/-/issues/202).

oscargus:
Closes #269 

(It was so quick, so I took a chance that it will be merged etc...)

I have no idea how to run to generate a VCD file though, but assuming that this is the sole file responsible for it and my search for locations didn't fail...

### DIFF
--- a/pymtl3/datatypes/PythonBits.py
+++ b/pymtl3/datatypes/PythonBits.py
@@ -529,3 +529,7 @@ class Bits:
   def hex( self ):
     str = "{:x}".format(int(self._uint)).zfill(((self._nbits-1)//4)+1)
     return "0x"+str
+
+  def bin_vcd(self):
+    str = "{:b}".format(int(self._uint)).zfill(self._nbits)
+    return str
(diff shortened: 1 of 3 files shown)


pymtl3/datatypes/PythonBits.py
```diff
@@ -529,3 +529,7 @@ def oct( self ):
   def hex( self ):
     str = "{:x}".format(int(self._uint)).zfill(((self._nbits-1)//4)+1)
     return "0x"+str
+
+  def bin_vcd(self):
```
cbatten, line 533: The method is named `bin_vcd` but then we call `vcd_bin` in the code below?

cbatten:
Looks like there might be a discrepancy between the method name and what is used to call that method. Let me try implementing these fixes in a branch in the pymtl3 repo.

cbatten:
I think this is now fixed on the main branch:

- https://github.com/pymtl/pymtl3/pull/292

@oscargus thanks for your help!

PR #292, opened 2025-08-22
Output VCD file compliant with the standard

cbatten:
PyMTL3 was generating an incorrect VCD format that just happened to work with GTKWave. The old version would output nets as:
```
b0b0 !
b0b0001 $
```
So every net had an extra `b0` at the beginning. In addition, single-bit nets should be represented by either 0 or 1 and then the symbol with no space. This pull request fixes this issue so that PyMTL3 now outputs:
```
0!
b0001 $
```


cbatten:
This fixes #269

